Issue #48, opened 2026-05-07
OAuth: redirect-URL allowlist uses HasPrefix without segment boundary


STGime:
**File:** `internal/tenant/auth_config.go:158-177`

The redirect-URL validator compares scheme+host exactly, then `strings.HasPrefix(candidate, allowedPrefix)`.

With allowlist entry `https://app.example.com/cb`, candidate `https://app.example.com/cb-evil` matches because there's no path-separator boundary.

## Concrete attack

If an attacker controls a path on the same host (e.g., user-content paths on a customer's domain, an open-redirect endpoint), they can intercept the OAuth code in the URL fragment.

## Fix

Require either exact path equality, or that the candidate path equals the allowed path or starts with `<allowed>/` (segment boundary). Reject if the next character after the prefix is not `/`, `?`, `#`, or end-of-string.

Severity: medium.

PR #111, opened 2026-05-12
Sec: auth.email() reads live from users table (closes #55)

### LINKED ISSUE #55: End-user JWT: email/role pulled from token, not re-checked against tenant DB
**Files:**
- `internal/auth/enduser_middleware.go:74-77`
- `internal/query/engine.go:48-54` (sets `app.end_user_email` GUC)

Email is taken from the JWT and stored as a GUC for RLS policies that may use `auth.email()`. If the user's email changes in the tenant DB after a token is issued, the GUC reflects the stale value until token expiry.

Tenant designs that gate access by email rather than ID will diverge from current state until the token rolls.

## Why it matters

Not a cross-tenant issue, but a per-tenant correctness/auth concern. Especially relevant if a user changes their email for security reasons (compromised contact, post-breach hygiene) — until token expiry, the old email still authenticates against email-keyed RLS.

## Fix

Either:
- Drop `app.end_user_email` and require RLS policies to look up the email from `users` keyed on `id` (canonical and current).
- Or add a token-revocation/version field bumped on email change.

Severity: low.

STGime:
Closes #55.

## Problem

\`auth.email()\` and per-tenant \`auth_email()\` returned the value of the \`app.end_user_email\` session GUC, populated from the JWT at request time. If a user changed their email in the tenant DB, the GUC still reflected the stale JWT value until token expiry — RLS policies gated by email diverged from the source of truth.

Severity: low (per the audit). But policies that email-gate sensitive ops (e.g. admin-by-email patterns) silently misbehave for the JWT TTL window.

## Fix

Migration 000052:
- Global \`auth.email()\` (added by #48) now does:
  \`SELECT email FROM users WHERE id = public.current_end_user_id()\`
  The unqualified \`users\` reference resolves via search_path to the caller's tenant schema (the engine sets that per request).
- Backfill loop rewrites each existing tenant's \`<schema>.auth_email()\` to the same shape (schema-qualified inside the format()).
- \`provision_tenant\` updated so freshly provisioned tenants are correct from creation.

Engine side: dropped the \`SET set_config('app.end_user_email', …)\` call — nothing reads the GUC anymore. The \`EndUserEmailFromContext\` getter stays since audit/logging still use the parsed-JWT email.

## Cost

One indexed PK lookup on \`users(id)\` per \`auth.email()\` evaluation. Acceptable for the correctness gain — policies gated by email now reflect current state instead of stale-token state.

## Test plan
- [x] \`go test ./...\` clean
- [ ] Smoke after deploy: sign in as Alice, change her email via \`/v1/auth/user\`, call any endpoint with an RLS policy that uses \`auth.email()\` — should match the new email immediately, not the old one.

### DIFF
--- a/internal/query/engine.go
+++ b/internal/query/engine.go
@@ -48,11 +48,11 @@ func (e *QueryEngine) applyRLSContext(ctx context.Context, tx pgx.Tx) error {
 		if _, err := tx.Exec(ctx, "SELECT set_config('app.end_user_role', 'authenticated', true)"); err != nil {
 			return fmt.Errorf("set end_user_role: %w", err)
 		}
-		if email := EndUserEmailFromContext(ctx); email != "" {
-			if _, err := tx.Exec(ctx, "SELECT set_config('app.end_user_email', $1, true)", email); err != nil {
-				return fmt.Errorf("set end_user_email: %w", err)
-			}
-		}
+		// app.end_user_email is no longer set — auth.email() and
+		// auth_email() now look up live from <schema>.users in
+		// migration 000052 (closes #55). The end-user email on the
+		// context is still used elsewhere (audit, logging) so the
+		// EndUserEmailFromContext getter stays.
 	}
 	return nil
 }
--- a/migrations/000052_auth_email_live_lookup.up.sql
+++ b/migrations/000052_auth_email_live_lookup.up.sql
@@ -17,8 +17,10 @@
 -- using `auth.email()` were already doing schema reads, so this is
 -- an extra row read per policy evaluation — acceptable for the
 -- correctness gain.
-
-BEGIN;
+--
+-- No explicit BEGIN/COMMIT: golang-migrate wraps each .up.sql in its
+-- own transaction by default, and an explicit COMMIT inside would
+-- commit the runner's outer tx prematurely (same footgun as #121).
 
 -- ── 1. Global auth.email() (added by 000048) ─────────────────────────
 -- search_path is set per-request to the tenant schema + public, so
@@ -279,5 +281,3 @@ END;
 $fn$;
 
 ALTER FUNCTION public.provision_tenant(UUID, TEXT, TEXT) OWNER TO eurobase_migrator;
-
-COMMIT;
--- a/migrations/000052_auth_email_live_lookup.up.sql
+++ b/migrations/000052_auth_email_live_lookup.up.sql
@@ -1,0 +1,283 @@
+-- 000052_auth_email_live_lookup.up.sql
+--
+-- Closes #55. The `auth.email()` and per-tenant `auth_email()` helpers
+-- used to read from the `app.end_user_email` session GUC, which the
+-- gateway populates from the JWT at request time. If a user changed
+-- their email in the tenant DB, the GUC still reflected the stale JWT
+-- value until token expiry — RLS policies gated by email diverged from
+-- the source of truth.
+--
+-- Both functions now read live from the current tenant schema's
+-- `users` table, keyed on `public.current_end_user_id()`. The lookup
+-- uses an unqualified `users` reference so search_path picks the
+-- right tenant schema; that path is set by the engine before every
+-- query (`SET LOCAL search_path TO %I, public`).
+--
+-- Per-row cost: one indexed PK lookup on `users(id)`. RLS policies
+-- using `auth.email()` were already doing schema reads, so this is
+-- an extra row read per policy evaluation — acceptable for the
+-- correctness gain.
+
+BEGIN;
+
+-- ── 1. Global auth.email() (added by 000048) ─────────────────────────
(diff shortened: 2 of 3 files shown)


migrations/000052_auth_email_live_lookup.up.sql
```diff
@@ -0,0 +1,283 @@
...
+-- right tenant schema; that path is set by the engine before every
+-- query (`SET LOCAL search_path TO %I, public`).
+--
+-- Per-row cost: one indexed PK lookup on `users(id)`. RLS policies
+-- using `auth.email()` were already doing schema reads, so this is
+-- an extra row read per policy evaluation — acceptable for the
+-- correctness gain.
+
+BEGIN;
+
```
STGime, line 24: **Potential issue (same as PR #121):** This migration uses explicit `BEGIN`/`COMMIT`. If golang-migrate wraps each `.up.sql` in its own transaction (which it does by default), the explicit `COMMIT` commits the runner's outer transaction prematurely. The down migration has the same pattern.

Safest fix: remove the explicit `BEGIN`/`COMMIT` and let golang-migrate handle the transaction. The `DO $$` block and `CREATE OR REPLACE` statements are all single-statement safe inside an implicit transaction.

---

migrations/000052_auth_email_live_lookup.up.sql
```diff
@@ -0,0 +1,283 @@
...
+-- ── 1. Global auth.email() (added by 000048) ─────────────────────────
+-- search_path is set per-request to the tenant schema + public, so
+-- the unqualified `users` reference resolves to the caller's tenant.
+CREATE OR REPLACE FUNCTION auth.email() RETURNS text
+    LANGUAGE sql STABLE AS $$
+    SELECT email FROM users WHERE id = public.current_end_user_id()
+$$;
+
+ALTER FUNCTION auth.email() OWNER TO eurobase_migrator;
```
STGime, line 33: The live lookup is the right call. `SELECT email FROM users WHERE id = public.current_end_user_id()` resolves `users` via search_path to the tenant schema — same mechanism the engine already uses for every tenant query. The PK lookup on `users(id)` is indexed and adds ~0.1ms per policy evaluation. Correct trade-off: one extra row read per `auth.email()` call vs silently diverging from the DB for the entire JWT TTL window.

---

internal/query/engine.go
```diff
@@ -48,11 +48,11 @@ func (e *QueryEngine) applyRLSContext(ctx context.Context, tx pgx.Tx) error {
...
-				return fmt.Errorf("set end_user_email: %w", err)
-			}
-		}
+		// app.end_user_email is no longer set — auth.email() and
+		// auth_email() now look up live from <schema>.users in
+		// migration 000052 (closes #55). The end-user email on the
+		// context is still used elsewhere (audit, logging) so the
+		// EndUserEmailFromContext getter stays.
 	}
 	return nil
```
STGime, line 57: Clean: the GUC write is dropped but `EndUserEmailFromContext` stays since audit/logging still use the JWT-parsed email. The comment explains why — prevents a future contributor from thinking it's dead code and removing the context propagation.

---

STGime (commented):
Correct fix for a real stale-data bug. The approach is sound: `auth.email()` reads live from `users(id)` via search_path instead of the stale JWT GUC. Cost is one indexed PK lookup per evaluation — negligible vs the correctness gain.

3 inline comments:

1. **`BEGIN`/`COMMIT` in migration (same issue as #121)** — golang-migrate auto-wraps, so the explicit `COMMIT` commits the runner's transaction prematurely. Remove the wrapping; every statement in this migration is safe in an implicit transaction.
2. **Live lookup design** — Correct. search_path resolution to the tenant schema, PK-indexed read, acceptable cost.
3. **GUC removal + context getter kept** — Clean. Comment explains the audit/logging dependency.

The `provision_tenant` rewrite is large but necessary — new tenants need the live-lookup `auth_email()` from creation. The down migration reverts cleanly to GUC-based behavior with a warning about stale-JWT reversion.

Fix the `BEGIN`/`COMMIT` wrapping before merge. Everything else is LGTM.

---

migrations/000052_auth_email_live_lookup.up.sql
STGime: Fixed in 98ec86e — removed the outer `BEGIN;`/`COMMIT;` from both up and down files. The inner `BEGIN`s inside `DO $$ ... $$` blocks (PL/pgSQL block delimiters, not SQL transactions) stay. Branch rebased on main; CI rerunning.

STGime (commented):
Feedback addressed. `BEGIN`/`COMMIT` removed from both `.up.sql` and `.down.sql`, with a comment explaining the golang-migrate transaction-wrapping footgun. Migration is now safe to run through the standard runner. LGTM — no remaining issues.

---

PR #113, opened 2026-05-12
Schedules: SDK control-plane for cron jobs (closes #112)

### LINKED ISSUE #112: Scheduled functions: SDK API for creating and managing cron jobs
## Problem

The platform supports cron jobs / scheduled function triggers, but the SDK
(`@eurobase/sdk@0.2.2`) currently only exposes `functions.invoke(name)` —
call an existing function. The actual cron schedule has to be configured
out-of-band in the dashboard, which means:

- The schedule isn't in the application repo, so it isn't versioned or
  reviewed in PRs.
- Reprovisioning a project from scratch requires the dashboard click-path
  to be remembered/documented separately from the function code it triggers.
- "Dashboard drift" — someone clicks the schedule off, no alert fires,
  and the cron silently stops working. We just hit this on LiveStylist
  while wiring `/internal/purge-images`; the workaround was Cloud
  Scheduler + a shared secret + a bash setup script. That's three moving
  parts to do what should be one function declaration.

## Proposed surface

Adding a `schedules` (or `functions.schedules`) sub-client. The minimum
useful shape — names are bikesheddable:

```ts
interface ScheduleSpec {
  /** Function to invoke. Must already be deployed. */
  functionName: string
  /** Standard 5-field cron expression, evaluated in `timezone`. */
  cron: string
  /** Optional IANA timezone. Defaults to UTC. */
  timezone?: string
  /** Optional JSON body / headers passed to the function on each tick. */
  payload?: unknown
  headers?: Record<string, string>
  /** Optional human-readable label that surfaces in the dashboard. */
  description?: string
  /** Whether the schedule is currently active. Defaults to true. */
  enabled?: boolean
}

interface ScheduleRow extends ScheduleSpec {
  id: string
  createdAt: string
  updatedAt: string
  lastRunAt?: string
  lastRunStatus?: 'success' | 'failure' | null
}

// On the SDK:
eb.functions.schedules.create(name: string, spec: ScheduleSpec):
  Promise<{ data: ScheduleRow | null; error: { status: number; message: string } | null }>
eb.functions.schedules.update(name: string, spec: Partial<ScheduleSpec>):
  Promise<{ data: ScheduleRow | null; error: ... }>
eb.functions.schedules.delete(name: string):
  Promise<{ error: ... }>
eb.functions.schedules.get(name: string):
  Promise<{ data: ScheduleRow | null; error: ... }>
eb.functions.schedules.list():
  Promise<{ data: ScheduleRow[]; error: ... }>
```

`name` is a stable identifier (e.g. `\"purge-expired-images\"`), separate
from the auto-generated `id`, so provisioning scripts can do
`create-or-update` idempotently without juggling UUIDs.

## Semantics worth nailing down up front

- **Idempotent provisioning.** Calling `create` for an existing name
  should either be an error (with an `already_exists` discriminator on
  the error) **or** a documented upsert. Picking either is fine; just
  pick one and document. I'd lean error + a `createOrUpdate` helper.
- **Catching up after downtime.** If the platform was down through a
  scheduled tick, does the next start backfill missed ticks (Quartz-style
  misfire policy) or just drop them? Cron jobs that depend on \"once per
  day max\" need the latter; jobs that depend on \"every tick eventually
  runs\" need the former. Most BaaS converge on \"drop missed\". Worth
  being explicit.
- **Per-execution timeout.** Default + override. Long-running purges
  shouldn't hold a worker forever.
- **Concurrency.** If a tick fires while the previous tick is still
  running, default to skipping. An option to queue/allow-overlap is nice
  but rarely needed.
- **Cron grammar.** Standard 5-field POSIX is fine; if you want to
  support `@daily` / `@hourly` aliases too, document them.
- **Auth.** Same `eb_sk_*` service key as the rest of the SDK — no need
  for a separate scope. Schedules don't carry user-level permissions.
- **Observability.** `lastRunAt` / `lastRunStatus` on the row is the
  cheapest useful telemetry; richer history (per-execution log link) can
  come later.

## What this would replace in an app's repo

Before:

```yaml
# cloudbuild.yaml
--set-secrets=...,PURGE_SECRET=purge-secret:latest

# scripts/setup-purge-scheduler.sh — 120 lines of gcloud + Secret Manager
# + Cloud Scheduler + IAM bindings, plus a /internal/purge-images endpoint
# in the backend that exists solely to be the target of the scheduler.
```

After:

```ts
// backend/scripts/provision-purge-job.ts
await eb.functions.schedules.create('purge-expired-images', {
  functionName: 'purge-expired-images',
  cron: '0 4 * * *',
  timezone: 'UTC',
  description: 'Daily purge of session_images rows past their 24h TTL',
})
```

…and the function itself becomes a one-line `DELETE` against the table.
No shared secret, no Cloud Scheduler, no external endpoint, no IAM
dance.

## Use case in flight

We're shipping the follow / feed feature on LiveStylist
([STGime/live-stylist-ai#1](https://github.com/STGime/live-stylist-ai/pull/1))
which adds `session_images` rows with a 24h TTL. The deletion job is the
prompt for this — currently set up via Cloud Scheduler as a stopgap.

Happy to be a beta tester / first caller once a draft API lands.

STGime:
## Summary

Closes #112. Adds `eb.functions.schedules.{create,update,get,list,delete,createOrUpdate}` so deployed edge functions can declare their own cron schedule in code instead of via the dashboard.

- Service-key (`eb_sk_*`) only — schedules are control-plane state.
- Mounted at `/v1/schedules` behind a new `requireSecretKeyForSchedules` gate.
- Worker's cron executor now invokes edge functions over the same HMAC-signed runner path the SDK uses; `shouldRun` evaluates cron in the schedule's timezone.

## Locked-in semantics (per #112 open questions)

- **Idempotency**: `create()` returns **409 / `already_exists`** on name collision (not upsert). `createOrUpdate()` is a separate helper for provisioning scripts.
- **Missed ticks**: dropped during downtime, no backfill.
- **Concurrency**: independent ticks; no overlap protection in v1.
- **Cron grammar**: POSIX 5-field only; no `@daily` / `@hourly` aliases.

## What's in the diff

**Server**
- `migrations/000053`: `function` action_type, `(project_id, name)` unique index, `timezone`/`payload`/`headers`/`description` columns.
- `internal/cron/service.go`: `GetByName`/`UpdateByName`/`DeleteByName`, `ErrNotFound` + `ErrNameAlreadyExists` sentinels.
- `internal/cron/executor.go`: new `function` action_type dispatch.
- `internal/cron/handler.go`: `SDKRoutes` using API-key project context.
- `internal/gateway/router.go`: mounts at `/v1/schedules`.
- `cmd/worker/main.go`: wires `FUNCTION_RUNNER_URL` + `FUNCTIONS_RUNNER_HMAC_SECRET` into the executor.

**SDK** (`@eurobase/sdk@0.3.0`)
- `src/schedules.ts` — `SchedulesClient` with discriminated `ScheduleError` codes.
- `src/functions.ts` — `eb.functions.schedules` sub-client.
- `test/schedules.mjs` — 8 tests over URL/payload shape, 409 → `already_exists`, name URL-encoding, `createOrUpdate` fall-through.

Migration renumbered to **000053** to avoid colliding with #111's open `000052_auth_email_live_lookup`.

## Test plan

- [x] \`go test ./...\` clean
- [x] \`npm run build && node test/schedules.mjs\` clean
- [ ] After merge: deploy succeeds, migration 000053 applies
- [ ] Smoke test from a tenant project: \`eb.functions.schedules.create('purge', { functionName: 'purge', cron: '0 4 * * *' })\` returns the row; a second call returns \`error.code === 'already_exists'\`.

### DIFF
--- a/internal/cron/executor.go
+++ b/internal/cron/executor.go
@@ -265,7 +265,11 @@ func (e *Executor) executeFunctionJob(ctx context.Context, job DueJob) error {
 	req.Header.Set("X-Schema-Name", job.SchemaName)
 	req.Header.Set("X-Function-Name", job.Action)
 	req.Header.Set("X-Function-ID", fnID)
-	req.Header.Set("X-Plan", "free") // schedule-fired invocations don't carry plan; runner uses for limits only
+	plan := job.Plan
+	if plan == "" {
+		plan = "free"
+	}
+	req.Header.Set("X-Plan", plan)
 	req.Header.Set("X-Cron-Job-ID", job.ID)
 	if req.Header.Get("Content-Type") == "" && len(job.Payload) > 0 {
 		req.Header.Set("Content-Type", "application/json")
@@ -283,6 +287,8 @@ func (e *Executor) executeFunctionJob(ctx context.Context, job DueJob) error {
 		respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
 		return fmt.Errorf("function returned %d: %s", resp.StatusCode, strings.TrimSpace(string(respBody)))
 	}
+	// Drain so the connection can be reused by http.Client's transport pool.
+	io.Copy(io.Discard, resp.Body)
 	return nil
 }
 
--- a/internal/cron/executor.go
+++ b/internal/cron/executor.go
@@ -1,39 +1,80 @@
 package cron
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"log/slog"
+	"net/http"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/eurobase/euroback/internal/functions"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 )
 
+// FunctionInvoker wires the cron executor to the function-runner HTTP
+// service so an `action_type='function'` schedule fires the named Deno
+// function with the schedule's `payload` and `headers`. Both fields are
+// optional — pass nil to disable function-typed schedules (they will be
+// recorded as failures with a clear message).
+type FunctionInvoker struct {
+	RunnerURL string
+	Signer    *functions.Signer
+	Client    *http.Client
+}
+
 // Executor runs due cron jobs on a schedule.
 type Executor struct {
-	svc  *CronService
-	pool *pgxpool.Pool
+	svc      *CronService
+	pool     *pgxpool.Pool
+	invoker  *FunctionInvoker
+	location *time.Location // fallback timezone when a job's tz fails to load
 }
 
 // NewExecutor creates a new cron job executor.
 func NewExecutor(svc *CronService, pool *pgxpool.Pool) *Executor {
-	return &Executor{svc: svc, pool: pool}
+	return &Executor{svc: svc, pool: pool, location: time.UTC}
+}
+
+// WithFunctionInvoker enables `action_type='function'` schedules by
+// configuring the HTTP path to the function runner. Pass a Signer from
+// the gateway/runner shared secret so the runner accepts the call. Safe
+// to call zero times — function schedules then fail-fast with a clear
+// message.
+func (e *Executor) WithFunctionInvoker(inv FunctionInvoker) *Executor {
+	if inv.Client == nil {
+		inv.Client = &http.Client{Timeout: 65 * time.Second}
+	}
+	e.invoker = &inv
+	return e
 }
 
 // RunDueJobs checks all enabled cron jobs and executes those whose schedule
-// matches the current time (truncated to the minute).
+// matches the current time (truncated to the minute, evaluated in the
+// job's configured timezone).
 func (e *Executor) RunDueJobs(ctx context.Context) error {
 	jobs, err := e.svc.GetDueJobs(ctx)
 	if err != nil {
 		return fmt.Errorf("get due jobs: %w", err)
 	}
 
-	now := time.Now().UTC()
+	now := time.Now()
 	for _, job := range jobs {
-		if !shouldRun(job.Schedule, now) {
+		loc := e.location
+		if job.Timezone != "" {
+			if l, err := time.LoadLocation(job.Timezone); err == nil {
+				loc = l
+			} else {
+				slog.Warn("cron job has invalid timezone, falling back to UTC",
+					"job_id", job.ID, "timezone", job.Timezone, "error", err)
+			}
+		}
+		if !shouldRun(job.Schedule, now.In(loc)) {
 			continue
 		}
 
@@ -135,6 +176,10 @@ func (e *Executor) RunDueJobs(ctx context.Context) error {
 //  4. Executes the action via `tx.Exec` — the extended query protocol,
 //     which runs only the first statement (defence-in-depth on top of
 //     the multi-statement rejection).
+//
+// `function` action_type is dispatched out-of-band to the Deno runner
+// over HTTP (same HMAC-signed path the SDK uses), so SQL-injection
+// concerns don't apply there — the action is the function name.
 func (e *Executor) executeJob(ctx context.Context, job DueJob) error {
 	switch job.ActionType {
 	case "sql":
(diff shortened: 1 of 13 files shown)


STGime (commented):
## Review of PR #113 — Schedules: SDK control-plane for cron jobs

Overall this is a well-structured PR. The security boundaries are solid (secret-key gating, reserved headers overwriting user-supplied ones, function name validation), the migration is clean, the SDK test suite is thorough, and the `enforcePlanLimit` refactor removes real duplication. A few items worth addressing before merge:

---

### 1. `X-Plan` hardcoded to `"free"` — **medium severity**

In `executor.go` → `executeFunctionJob`:

```go
req.Header.Set("X-Plan", "free") // schedule-fired invocations don't carry plan; runner uses for limits only
```

This means **every** schedule-fired function invocation hits the runner with free-tier limits, even for pro-plan projects. If the runner enforces execution time, memory, or concurrency caps based on `X-Plan`, pro customers will be silently degraded. The project's actual plan is one query away (the `GetDueJobs` query already JOINs `projects` — you could add `p.plan` to the SELECT and carry it on `DueJob`).

### 2. Response body not drained on success — **low severity**

In `executeFunctionJob`, on a successful response (`< 400`), the body is closed via `defer` but never read/drained. This prevents HTTP connection reuse in Go's `http.Client` transport pool. For a cron executor that may fire many functions per tick, this adds unnecessary TCP overhead. A one-liner fix:

```go
io.Copy(io.Discard, resp.Body)
```

before the return.

### 3. No request body size limit on SDK handlers — **low severity**

The SDK handlers (`handleSDKCreate`, `handleSDKUpdate`) call `json.NewDecoder(r.Body).Decode(...)` without a size limit. A malicious holder of a secret key could POST a multi-GB body. Consider wrapping with `http.MaxBytesReader`:

```go
r.Body = http.MaxBytesReader(w, r.Body, 1<<20) // 1 MB
```

This may already be handled at a higher middleware layer — if so, disregard.

### 4. `createOrUpdate` race window — **informational**

The SDK's `createOrUpdate` does create-then-update-on-409. Between the 409 and the PATCH, another caller could delete the schedule, causing the update to 404. This is fine for a provisioning helper (documented as such), just noting the window exists. A server-side upsert endpoint would close it if it ever matters.

### 5. Minor nits

- **`scanCronJob` reuse**: Nice pattern extracting the scan into a shared helper — eliminates the column-list drift that was happening before.
- **`nullableJSON`**: Clean way to make `COALESCE` work with JSONB optionals.
- **Test coverage**: 8 SDK tests with a stubbed fetch is solid. The `createOrUpdate` fall-through test (test 8) with a multi-call mock is well done.
- **Migration down.sql**: Correctly restores the `action_type` CHECK constraint to `('sql', 'rpc')` only. Good.

---

**Verdict**: Items 1 is worth fixing before merge (pro customers hitting free limits is a real bug). Items 2-3 are nice-to-haves. The rest is clean — good work.

---

internal/cron/executor.go
```diff
@@ -158,11 +203,89 @@ func (e *Executor) executeJob(ctx context.Context, job DueJob) error {
...
+	// Caller-supplied headers go on first; reserved identity headers
+	// overwrite so a malicious schedule can't spoof project_id.
+	if len(job.Headers) > 0 {
+		var hdrs map[string]string
+		if err := json.Unmarshal(job.Headers, &hdrs); err == nil {
+			for k, v := range hdrs {
+				req.Header.Set(k, v)
+			}
+		}
+	}
```
STGime, line 263: **Bug (medium):** This hardcodes every schedule-fired invocation to free-tier runner limits, even for pro-plan projects. If the runner enforces execution time, memory, or concurrency caps based on `X-Plan`, pro customers will be silently degraded.

`GetDueJobs` already JOINs `projects` — add `p.plan` to the SELECT, carry it on `DueJob`, and use it here:

```go
plan := job.Plan
if plan == "" {
    plan = "free"
}
req.Header.Set("X-Plan", plan)
```

---

internal/cron/executor.go
```diff
@@ -158,11 +203,89 @@ func (e *Executor) executeJob(ctx context.Context, job DueJob) error {
...
+	req.Header.Set("X-Plan", "free") // schedule-fired invocations don't carry plan; runner uses for limits only
+	req.Header.Set("X-Cron-Job-ID", job.ID)
+	if req.Header.Get("Content-Type") == "" && len(job.Payload) > 0 {
+		req.Header.Set("Content-Type", "application/json")
+	}
+	if e.invoker.Signer != nil {
+		e.invoker.Signer.Sign(req.Header, time.Now())
+	}
+
+	resp, err := e.invoker.Client.Do(req)
```
STGime, line 281: **Nit (low):** On success the response body is closed via `defer` but never drained, which prevents HTTP connection reuse in Go's `http.Client` transport pool. For a cron executor that may fire many functions per tick, this adds unnecessary TCP overhead. Consider adding before the final `return nil`:

```go
io.Copy(io.Discard, resp.Body)
```

---

internal/cron/handler.go
```diff
@@ -155,6 +160,169 @@ func handleListRuns(svc *CronService) http.HandlerFunc {
...
+			jsonError(w, "internal server error", http.StatusInternalServerError)
+			return
+		}
+		jsonResponse(w, job, http.StatusOK)
+	}
+}
+
+func handleSDKCreate(svc *CronService) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		projectID, ok := projectIDFromAPIKey(r)
```
STGime, line 218: **Suggestion (low):** The SDK handlers decode `r.Body` without a size limit. A malicious holder of a secret key could POST an arbitrarily large body. Consider wrapping with `http.MaxBytesReader` in `handleSDKCreate` and `handleSDKUpdate`:

```go
r.Body = http.MaxBytesReader(w, r.Body, 1<<20) // 1 MB
```

If there's already a global body-size middleware upstream, disregard.

---

sdk/js/src/schedules.ts
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Idempotent: create the schedule if missing, update it if present.
+   * Useful for provisioning scripts that want a single declaration.
+   */
+  async createOrUpdate(
+    name: string,
+    spec: ScheduleSpec,
+  ): Promise<{ data: ScheduleRow | null; error: ScheduleError | null }> {
+    const created = await this.create(name, spec)
+    if (!created.error) return created
```
STGime, line 242: **Informational:** Small race window — between the 409 from `create()` and the subsequent `update()`, another caller could delete the schedule, causing the update to 404. Totally fine for a provisioning helper (the docstring makes the intent clear), just noting the window exists. A server-side `INSERT ... ON CONFLICT DO UPDATE` endpoint would close it if it ever matters in practice.

---

STGime (commented):
4 inline comments added. The `X-Plan` hardcode (executor.go:263) is the one I'd fix before merge — pro customers hitting free-tier limits is a real bug. The rest are low-severity suggestions.

---

Issue #70, opened 2026-05-07
Selecting Pro at project creation persists plan as "free"


STGime:
**Reported behaviour:** User selected Pro when creating a project. The project was created but stored with `plan='free'`.

## Confirmed evidence

**DB row** (after creation):
\`\`\`
ID         NAME      PLAN  CREATED_AT
1ac63ce5…  superapp  free  2026-05-07T11:21:50.768272Z
\`\`\`

**Network request body** (captured in browser devtools, separate repro with project name "ddof"):
\`\`\`json
{ "name": "ddof", "plan": "pro", "region": "fr-par", "slug": "ddof" }
\`\`\`

So **the frontend correctly sends `plan: "pro"`** — the bug is between the HTTP request and the DB INSERT.

## What I traced

1. `internal/tenant/handler.go:124-126`: defaults `req.Plan` to `"free"` only if empty; otherwise passes through.
2. `internal/tenant/service.go:114-118`: `INSERT INTO projects (..., plan) VALUES (..., $7)` with `req.Plan`. No CHECK constraint, no trigger that rewrites.
3. `provision_tenant()` doesn't touch `projects.plan` (only sets `schema_name`).

So the JSON body says `"plan":"pro"` but the row ends up `'free'`. Possible mechanisms not yet ruled out:

- **Limits-service downgrade**: `CheckProjectLimit` (line ~85 of handler.go) might silently force-downgrade pro → free if the user has no active billing subscription. Worth checking `internal/plans/`.
- **JSON decoding dropping the field**: unlikely (\`Plan string \\\`json:"plan"\\\`\` should decode), but worth printing `req.Plan` right after `json.NewDecoder(...).Decode(&req)` to confirm.
- **A second handler** registered at the same path overriding behaviour. Unlikely with chi but worth a check.

## Quickest debug

Add `slog.Info("create project request", "plan", req.Plan, "name", req.Name)` right after `Decode(&req)` in `handler.go:94`. Repro with Pro selected. The log will show whether `req.Plan` is `"pro"` or `"free"` at that point — pinpoints whether the bug is before or after JSON decoding.

## Side bug noticed during investigation

`console/src/routes/(app)/projects/+page.svelte:55` sends `region: 'eu-west-paris'`. The backend (`internal/tenant/handler.go:118`) only accepts `'fr-par'`. So the projects-page modal's "Create New Project" button should 400 with `region must be fr-par`. The onboarding page is correct. (Independent of this issue — file separately if not already covered.)

[...]

PR #114, opened 2026-05-12
#70: observability for plan persistence + regression test + side-bug fix

STGime:
Triage for #70 — \"Selecting Pro at project creation persists plan as free.\"

## What I traced

Every path that touches `projects.plan`:

- Handler defaults only when `req.Plan == \"\"` — frontend always sends.
- Service `INSERT INTO projects (..., plan) VALUES (..., \$7)` with `req.Plan` directly.
- `provision_tenant(p_plan)` declares the parameter but never uses it.
- No CHECK, no trigger, no subsequent `UPDATE plan` anywhere in the codebase.
- `CheckProjectLimit` is pure read; doesn't downgrade.
- No Mollie/subscription code (deferred per CLAUDE.md).

Net: the reported bug is **not reproducible from the code**. Rather than guess a fix, this PR adds the instrumentation needed to pin the next repro.

## What lands

1. **Two diagnostic `slog.Info` calls** in `HandleCreateProject`: one right after JSON decode, one right before the service call. The next time someone selects Pro and the row ends up free, the logs show whether `req.Plan` was `\"pro\"` or `\"free\"` at each point — pinning the bug to the HTTP body, the JSON decode, or somewhere downstream. Annotated with `#70 diagnostic` so they're easy to remove once closed.

2. **`TestCreateProject_PlanPersisted`** — table-driven regression test over `free` / `pro` that reads the plan back **from the DB row** (not the returned struct, which echoes `req.Plan` and would have hidden the bug). Integration-test style, skips in `-short` mode.

3. **Side-bug fix** the issue noted in passing: `console/src/routes/(app)/projects/+page.svelte` sent `region: 'eu-west-paris'` which the backend rejects (only `fr-par` is accepted). Modal now matches the onboarding flow.

## Test plan

- [ ] Merge + deploy
- [ ] Reproduce: create a project, select Pro
- [ ] `kubectl logs deployment/gateway -n eurobase | grep \"create project\"` — note the `plan=` values in both log lines
- [ ] If both say `pro`: bug is downstream of the service call (something in `provision_tenant` or a constraint I missed). File follow-up.
- [ ] If first says `pro`, second says `free`: bug is in some middleware between decode and service. Unlikely but possible.
- [ ] If first says `free` (request body claims pro but server saw free): bug is in browser→server transit. Check service worker, CORS, content-type encoding.
- [ ] Workaround for affected projects: `UPDATE projects SET plan='pro' WHERE id=…`

### DIFF
--- a/console/src/routes/(app)/projects/+page.svelte
+++ b/console/src/routes/(app)/projects/+page.svelte
@@ -52,7 +52,7 @@
 			await api.createProject({
 				name: newName.trim(),
 				slug: newSlug,
-				region: 'eu-west-paris',
+				region: 'fr-par',
 				plan: newPlan
 			});
 			showNewModal = false;
--- a/internal/tenant/service_test.go
+++ b/internal/tenant/service_test.go
@@ -95,6 +95,42 @@ func insertTestPlatformUser(t *testing.T, pool *pgxpool.Pool, email string) stri
 	return id
 }
 
+// TestCreateProject_PlanPersisted is the regression test for #70.
+// Asserts that a Pro plan in the request body survives the INSERT and is
+// readable back from the DB row — not just echoed from the in-memory req.
+func TestCreateProject_PlanPersisted(t *testing.T) {
+	pool := setupTestDB(t)
+	ctx := context.Background()
+	svc := &TenantService{pool: pool}
+
+	for _, plan := range []string{"free", "pro"} {
+		t.Run(plan, func(t *testing.T) {
+			uid := insertTestPlatformUser(t, pool, "plan-"+plan+"@test.eurobase.local")
+			project, err := svc.CreateProject(ctx, uid, "plan-"+plan+"@test.eurobase.local", CreateProjectRequest{
+				Name:   "Plan " + plan,
+				Slug:   "plan-test-" + plan,
+				Region: "fr-par",
+				Plan:   plan,
+			})
+			if err != nil {
+				t.Fatalf("CreateProject(plan=%q): %v", plan, err)
+			}
+			t.Cleanup(func() { cleanupProject(t, pool, project.ID) })
+
+			// Read back from the DB — don't trust the returned struct
+			// since it echoes req.Plan rather than re-SELECTing.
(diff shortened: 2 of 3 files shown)


console/src/routes/(app)/projects/+page.svelte
```diff
@@ -52,7 +52,7 @@
 			await api.createProject({
 				name: newName.trim(),
 				slug: newSlug,
-				region: 'eu-west-paris',
+				region: 'fr-par',
```
STGime, line 55: Good catch — `eu-west-paris` would have hit the region validation and returned a 400. This was likely the actual cause of #70: the modal's `createProject` call fails silently or the error is swallowed, and the user sees the project list without the new project (or a fallback path creates it with defaults). Worth checking if there's an error handler around this `api.createProject` call that might mask the failure.

---

internal/tenant/service_test.go
```diff
@@ -95,6 +95,42 @@ func insertTestPlatformUser(t *testing.T, pool *pgxpool.Pool, email string) stri
...
+				Slug:   "plan-test-" + plan,
+				Region: "fr-par",
+				Plan:   plan,
+			})
+			if err != nil {
+				t.Fatalf("CreateProject(plan=%q): %v", plan, err)
+			}
+			t.Cleanup(func() { cleanupProject(t, pool, project.ID) })
+
+			// Read back from the DB — don't trust the returned struct
```
STGime, line 120: Smart — reading from the DB row instead of the returned struct catches the real bug surface. If `CreateProject` ever starts echoing a stale value while the INSERT wrote something different, this catches it.

---

STGime (commented):
Clean, methodical triage PR. The diagnostic logs are well-annotated (`#70 diagnostic`) for easy removal, the regression test reads from the DB row (not the returned struct — correct), and the `eu-west-paris` → `fr-par` fix is a real side-bug that may well be the root cause of #70 itself (the request would 400 with the wrong region, and depending on error handling in the modal, the user might never see it).

No blockers — LGTM.

---

PR #119, opened 2026-05-13
Migration History: stop stamping backfill rows with now()

STGime:
## Bug

Opening **Database → Migration History** after any out-of-band DDL (SQL runner, MCP \`runSQL\`, etc.) makes pre-existing tables appear as freshly \"Created today\". Reported via LiveStylist: a single \`ALTER TABLE app_users ADD COLUMN stable_device_id\` produced four phantom \"Created table\" rows (\`blocks\`, \`vault_secrets\`, \`session_images\`, \`follows\`) stamped at the exact minute the page was opened.

## Root cause

\`backfillUnloggedTables\` (\`internal/query/ddl_handler.go:207\`) runs on every page load. For each table in \`information_schema\` with no matching log entry, it inserts a synthetic \`create_table\` row — but omits \`created_at\`, so the column defaults to \`now()\`. Every page open fabricates audit-trail entries with the visit timestamp.

The \`detail\` JSON already carries \`{\"source\":\"backfill\"}\`, but the UI ignores it and just renders \`created_at\`.

## Fix

- **Backend** — backfill INSERT now explicitly sets \`created_at = NULL\`. \`SchemaChange.CreatedAt\` is \`*time.Time\` so NULL round-trips. Pre-existing tables sort to the bottom under \`ORDER BY created_at DESC\` (Postgres NULLS LAST default for DESC).
- **Frontend** — \`history/+page.svelte\` renders rows where \`detail.source === 'backfill'\` (or \`created_at === null\`) as **\"Detected table … existed before tracking\"** with a tooltip explaining the discovery-vs-event distinction.
- **Migration 000054** — NULLs out \`created_at\` on backfill rows already in the DB so deployed projects render correctly immediately, not just for fresh discoveries.

## Out of scope

DDL that lands via SQL runner / MCP \`runSQL\` (like the actual \`ALTER TABLE app_users\` that triggered this report) still bypasses the \`schema_changes\` hook entirely — the backfill only stub-detects tables, never columns or alters. That's a separate gap worth filing if it bites.

## Test plan

- [x] \`go build ./... && go test ./internal/query/...\` clean
- [ ] Merge + deploy
- [ ] Open LiveStylist → Database → Migration History → the four 13 May 21:26 entries now read \"Detected table … existed before tracking\" (no timestamp)
- [ ] Create a real table via the console → still shows \"Created table\" with the correct timestamp

### DIFF
--- a/internal/query/ddl_handler.go
+++ b/internal/query/ddl_handler.go
@@ -43,16 +43,19 @@ type AddColumnRequest struct {
 	DefaultValue string `json:"default_value,omitempty"`
 }
 
-// SchemaChange represents a logged DDL operation.
+// SchemaChange represents a logged DDL operation. CreatedAt is nullable
+// because backfill rows (synthetic "we discovered this table existed"
+// entries) have no real creation timestamp; the column carries NULL for
+// those and the console UI renders them as "Detected".
 type SchemaChange struct {
-	ID         string    `json:"id"`
-	ProjectID  string    `json:"project_id"`
-	Action     string    `json:"action"`
-	TableName  string    `json:"table_name"`
-	ColumnName *string   `json:"column_name"`
-	Detail     any       `json:"detail"`
-	SQLText    *string   `json:"sql_text"`
-	CreatedAt  time.Time `json:"created_at"`
+	ID         string     `json:"id"`
+	ProjectID  string     `json:"project_id"`
+	Action     string     `json:"action"`
+	TableName  string     `json:"table_name"`
+	ColumnName *string    `json:"column_name"`
+	Detail     any        `json:"detail"`
+	SQLText    *string    `json:"sql_text"`
+	CreatedAt  *time.Time `json:"created_at"`
 }
 
 // RenameTableRequest is the JSON body for PATCH /schema/tables/{table}.
@@ -230,10 +233,16 @@ func backfillUnloggedTables(ctx context.Context, pool *pgxpool.Pool, projectID, 
 		if platformTables[tableName] {
 			continue
 		}
-		// Insert a backfilled create_table entry.
+		// Insert a backfilled entry with NULL created_at — these rows
+		// represent tables we *discovered* in information_schema, not
+		// tables we observed being created. Defaulting created_at to
+		// now() makes every page load look like a fresh DDL event,
+		// which is the bug. Frontend renders NULL as "Detected
+		// (existed before tracking)" with no specific timestamp; NULL
+		// also sorts to the bottom under `ORDER BY created_at DESC`.
 		_, _ = pool.Exec(ctx,
-			`INSERT INTO schema_changes (project_id, action, table_name, detail)
-			 VALUES ($1, 'create_table', $2, '{"source":"backfill"}'::jsonb)`,
+			`INSERT INTO schema_changes (project_id, action, table_name, detail, created_at)
+			 VALUES ($1, 'create_table', $2, '{"source":"backfill"}'::jsonb, NULL)`,
 			projectID, tableName)
 		slog.Info("backfilled schema change", "project_id", projectID, "table", tableName)
 	}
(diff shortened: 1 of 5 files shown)


migrations/000054_schema_changes_null_backfill_timestamps.up.sql
```diff
@@ -0,0 +1,21 @@
...
+-- as "Detected (existed before tracking)" instead of pretending to
+-- have been created moments ago.
+--
+-- Identifies backfill rows via the existing detail marker
+-- (`{"source":"backfill"}`) — that flag has been written since the
+-- backfill was introduced, so this is precise and reversible.
+
+UPDATE public.schema_changes
+SET created_at = NULL
+WHERE detail->>'source' = 'backfill';
```
STGime, line 21: **Potential blocker:** If `schema_changes.created_at` was defined with `NOT NULL DEFAULT now()` (which is the common pattern and would explain why the old INSERT worked — it omitted the column and got the default), then both this migration (`SET created_at = NULL`) and the new explicit `NULL` INSERT in `ddl_handler.go` will fail at runtime with a NOT NULL violation.

You'd need to add this before the UPDATE:

```sql
ALTER TABLE public.schema_changes ALTER COLUMN created_at DROP NOT NULL;
```

If the column already allows NULL, disregard — but worth verifying against the original `CREATE TABLE` / migration for `schema_changes`.

---

migrations/000054_schema_changes_null_backfill_timestamps.down.sql
```diff
@@ -0,0 +1,10 @@
+-- Restore a timestamp on backfill rows. We can't recover the original
+-- (since there never was one — the rows were synthesised), but using
+-- `to_timestamp(0)` (1970-01-01) at least keeps the column non-NULL if
+-- something downstream relied on that, while making the "this is fake"
+-- nature obvious.
+
+UPDATE public.schema_changes
+SET created_at = to_timestamp(0)
```
STGime, line 8: **Nit:** `to_timestamp(0)` for the down migration is a nice choice — obviously fake (1970-01-01) so it won't be confused with a real event, but keeps the column non-NULL for anything that relied on it. Good.

---

internal/query/ddl_handler.go
```diff
@@ -43,16 +43,19 @@ type AddColumnRequest struct {
...
-	ProjectID  string    `json:"project_id"`
-	Action     string    `json:"action"`
-	TableName  string    `json:"table_name"`
-	ColumnName *string   `json:"column_name"`
-	Detail     any       `json:"detail"`
-	SQLText    *string   `json:"sql_text"`
-	CreatedAt  time.Time `json:"created_at"`
+	ID         string     `json:"id"`
+	ProjectID  string     `json:"project_id"`
+	Action     string     `json:"action"`
```
STGime, line 53: **Safety check:** With `CreatedAt` now `*time.Time`, any code that previously accessed `change.CreatedAt` without a nil check will panic. The JSON serialization is fine (`null`), but verify there are no other consumers of `SchemaChange` that format or compare `.CreatedAt` directly — e.g. audit log helpers, sorting functions, or template rendering outside this file.

---

STGime (commented):
Good bug find and clean fix. The root cause analysis is spot-on — `backfillUnloggedTables` defaulting `created_at` to `now()` on every page load is exactly the kind of subtle data corruption that's hard to notice until it's everywhere.

3 inline comments:

1. **Migration may fail if `created_at` is `NOT NULL` (potential blocker)** — The old INSERT worked by omitting the column (getting the default), but the new explicit `NULL` and the `UPDATE SET created_at = NULL` will both violate a NOT NULL constraint. Verify the column definition; if it's `NOT NULL DEFAULT now()`, add `ALTER COLUMN created_at DROP NOT NULL` to the migration.

2. **`*time.Time` nil safety** — Any existing code that accesses `SchemaChange.CreatedAt` without a nil check will panic. JSON is fine (`null`), but verify no other consumers format or compare it directly.

3. Down migration using `to_timestamp(0)` is a nice touch — obviously synthetic, won't confuse with real events.

Frontend changes are clean — the `isBackfill` helper checking both `detail.source` and `created_at === null` is good belt-and-suspenders, and the tooltip explaining discovery-vs-event is the right UX.

---

STGime:
Superseded by #121, which deletes the buggy backfill entirely and audits all DDL via Postgres event triggers. The mis-stamping symptom #119 fixed is no longer possible since there are no backfilled rows being written with now() anymore.

PR #121, opened 2026-05-13
Migration History: audit ALL DDL via Postgres event triggers (closes #120)

### LINKED ISSUE #120: Migration History: track DDL from all sources (SQL runner, MCP, direct), not just the platform DDL hook
## Problem

Today only DDL that flows through the platform's typed DDL handlers (\`HandleCreateTable\`, \`HandleAddColumn\`, etc.) reaches \`schema_changes\` via \`logSchemaChange\`. Every other path silently bypasses the audit:

- **Console SQL runner** (\`/platform/projects/{id}/data/sql\` → \`query.HandlePlatformSQL\`) — runs arbitrary SQL, no logging hook.
- **Console SQL runner transaction** (\`/data/sql/transaction\`) — same.
- **MCP \`runSQL\` / \`runSQLTransaction\`** — same.
- **SDK \`/v1/db/sql\`** — same.
- **Anything done via direct DB access** — out of platform control entirely.

Surfaced by today's #119 postmortem: an \`ALTER TABLE app_users ADD COLUMN stable_device_id\` issued via MCP \`runSQL\` does not appear in Migration History at all. PR #119 fixed the *backfill* mis-stamping, but the underlying gap — that the migration history is a partial ledger — remains.

## Acceptance

**\"ALL migrations, regardless of source, should be in the list.\"** — every CREATE / ALTER / DROP / RENAME / INDEX-ish change on tenant schemas, regardless of whether it was issued via the console SQL editor, MCP, the SDK's \`runSQL\`, or anything else, should show up in Migration History with an accurate timestamp.

## Approach (suggested)

**A. Postgres event trigger** — the right answer.

Event triggers fire on \`ddl_command_end\` regardless of how the command reached Postgres. A trigger function in \`public.*\` (owned by \`eurobase_migrator\`, deployed via a migration) walks \`pg_event_trigger_ddl_commands()\`, derives the project_id from \`schema_name\` → \`projects\` lookup, and inserts the appropriate \`schema_changes\` row. One trigger covers every code path forever.

Pros:
- Catches direct DB access too — defense in depth for the audit log.
- No code change in the SQL runner / MCP server / SDK; the audit is at the DB layer.
- Survives future refactors of the DDL handlers.

Cons:
- Event triggers are global per database; the function has to filter to tenant schemas (skip \`public.*\` migrations, skip platform-owned tables).
- Has to handle \`CREATE TABLE … LIKE\`, partitions, inherited tables, and other edge cases the typed handlers never see.
- Owner mismatch: the trigger runs as the calling role; needs SECURITY DEFINER to insert into \`schema_changes\` if the caller (e.g. \`eurobase_gateway\`) doesn't have INSERT on that table.

**B. Wrapper around the SQL runners** — partial fix.

\`query.HandlePlatformSQL\` / \`HandleSQL\` could parse the executed SQL post-hoc and call \`logSchemaChange\` for any DDL-shaped statement. Simpler to implement but:
- Re-implements DDL parsing in Go (fragile).
- Doesn't cover direct DB access.
- Doesn't cover SDK DDL paths that don't use \`HandleSQL\` (e.g. \`HandleDDL\` on \`/v1/db/schema/tables\`).

Recommend **A**. \`B\` is a stopgap if A is too much for a sprint.

## Notes

- The existing backfill (now NULL-stamped per #119) becomes mostly redundant once the event trigger is in place — newly-created tables get a real timestamp. The backfill stays useful for tables that existed *before* the event trigger was installed.
- The trigger has to be installed in every tenant DB. Since Eurobase is single-DB multi-schema, that's once globally. Migration \`000055_*.up.sql\` (or whatever slot is free) ships it.
- \`logSchemaChange\` from the typed handlers can stay as-is — it'd write duplicate rows for the same DDL (one from the handler, one from the event trigger). Either deduplicate (PRIMARY KEY constraint won't help; the rows have different ids) or have the typed handlers stop logging and let the trigger be the single source of truth. Probably the latter — single source is cleaner.

## Severity

Medium. Real impact for any tenant who uses the SQL editor: they make changes, Migration History claims nothing happened. Confusing for compliance / audit conversations. Low-risk to fix; high-trust gain.

STGime:
Closes #120. Today only DDL that flows through the platform's typed handlers reaches \`schema_changes\`; SQL runner / MCP \`runSQL\` / SDK \`runSQL\` / direct DB access all bypass logging. Migration History is a partial ledger.

### Migration 000055 — two Postgres event triggers

- \`trg_log_ddl_end\` ON \`ddl_command_end\` → CREATE / ALTER
- \`trg_log_ddl_drop\` ON \`sql_drop\` → DROP

Both functions are SECURITY DEFINER, owned by \`eurobase_migrator\`, filter to tenant schemas (\`tenant_*\`) only, and skip the platform's own DDL on \`public.*\`. Inserts are wrapped in EXCEPTION blocks so an audit failure cannot abort the user's DDL.

Mapping:

| Postgres event | schema_changes.action |
|---|---|
| \`CREATE TABLE\` | \`create_table\` (with column list in detail) |
| \`CREATE INDEX\` | \`create_index\` (target table resolved via pg_index) |
| \`ALTER TABLE\` on \`table column\` | \`add_column\` / \`alter_column\` (disambiguated by sniffing \`current_query()\` for \`\\yADD\\s+COLUMN\\y\`) |
| \`ALTER TABLE … RENAME TO\` | \`rename_table\` |
| \`DROP TABLE\` / \`DROP COLUMN\` / \`DROP INDEX\` (via sql_drop) | \`drop_table\` / \`drop_column\` / \`drop_index\` |

### Code

- \`internal/query/ddl_handler.go\`:
  - **Removed** \`backfillUnloggedTables\` entirely — the event trigger eliminates the need to synthesise \`create_table\` rows from \`information_schema\`, and the backfill was the source of the #119 mis-stamping bug.
  - \`logSchemaChange\` no longer writes to \`schema_changes\` (the event trigger owns that table now). It still writes to \`audit_log\` so DDL appears in the audit feed.
  - \`SchemaChange.CreatedAt\` → \`*time.Time\` so legacy NULL rows round-trip.
- Console \`history/+page.svelte\` + \`api.ts\`:
  - NULL \`created_at\` renders as **\"existed before tracking\"** with a tooltip explaining the discovery-vs-event distinction.
  - \`actionLabel\` expanded to cover the new action values the trigger emits (\`alter_column\`, \`rename_table\`, \`create_index\`, \`drop_index\`, FK / unique constraint variants).

## Relationship with #119

This PR conceptually supersedes #119. Either order lands cleanly:

- **#119 first**: the frontend changes overlap (both wire NULL-aware rendering); the merge picks up cleanly because the edits are equivalent.
- **This PR first**: #119 becomes unnecessary and can be closed — backfill is gone, no mis-stamping path remains.

## Out of scope

The trigger does the \`pg_attribute\` lookup to enrich the detail JSON for column events, but \`formatDetail\` in the history page expects the typed-handler shape (rich \`{type, nullable, default}\` for ADD COLUMN, \`{columns: […]}\` for CREATE TABLE). Bringing the trigger output in line with the typed-handler detail shape is a follow-up.

## Test plan

- [x] \`go build ./... && go test ./internal/query/... ./internal/gateway/...\` clean
- [ ] Merge + deploy → migration 000055 applies → event triggers installed cluster-wide
- [ ] On LiveStylist: run \`ALTER TABLE app_users ADD COLUMN test_col text\` via MCP \`runSQL\`; open Migration History → entry appears with the correct timestamp
- [ ] Run \`DROP TABLE\` via SQL editor on a throwaway table → \`drop_table\` row appears
- [ ] Trigger an obvious failure path in the trigger function (e.g. malformed projects row) and verify the user's DDL still commits

### DIFF
--- a/internal/query/ddl_handler.go
+++ b/internal/query/ddl_handler.go
@@ -1,7 +1,6 @@
 package query
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"log/slog"
@@ -43,16 +42,19 @@ type AddColumnRequest struct {
 	DefaultValue string `json:"default_value,omitempty"`
 }
 
-// SchemaChange represents a logged DDL operation.
+// SchemaChange represents a logged DDL operation. CreatedAt is nullable
+// for historical backfill rows that pre-date the event-trigger audit
+// (#119/#120). Going forward every row written by the event triggers
+// carries a real timestamp.
 type SchemaChange struct {
-	ID         string    `json:"id"`
-	ProjectID  string    `json:"project_id"`
-	Action     string    `json:"action"`
-	TableName  string    `json:"table_name"`
-	ColumnName *string   `json:"column_name"`
-	Detail     any       `json:"detail"`
-	SQLText    *string   `json:"sql_text"`
-	CreatedAt  time.Time `json:"created_at"`
+	ID         string     `json:"id"`
+	ProjectID  string     `json:"project_id"`
+	Action     string     `json:"action"`
+	TableName  string     `json:"table_name"`
+	ColumnName *string    `json:"column_name"`
+	Detail     any        `json:"detail"`
+	SQLText    *string    `json:"sql_text"`
+	CreatedAt  *time.Time `json:"created_at"`
 }
 
 // RenameTableRequest is the JSON body for PATCH /schema/tables/{table}.
@@ -153,23 +155,15 @@ func HandleRLSAudit(pool *pgxpool.Pool) http.HandlerFunc {
 // HandleSchemaChanges returns a handler that lists schema change history.
 // GET /platform/projects/{id}/schema/changes
 //
-// On each request it also backfills any tables that exist in the tenant
-// schema but have no "create_table" entry — this catches tables created
-// via CLI migrations, psql, or any path that bypasses the DDL handlers.
+// Closes #120: every CREATE / ALTER / DROP on a tenant schema is logged by
+// the public.log_ddl_event() / log_ddl_drop() Postgres event triggers
+// (migration 000055), regardless of which path issued the DDL (typed
+// handlers, SQL runner, MCP, SDK, direct DB access). This handler is a
+// straight read against schema_changes — no backfill, no client-side
+// reconciliation.
 func HandleSchemaChanges(pool *pgxpool.Pool) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		projectID := chi.URLParam(r, "id")
-
-		// Resolve tenant schema for backfill.
-		var schemaName string
-		_ = pool.QueryRow(r.Context(),
-			`SELECT schema_name FROM projects WHERE id = $1`, projectID,
-		).Scan(&schemaName)
-
-		// Backfill: find tables in the schema that have no create_table log entry.
-		if schemaName != "" {
-			backfillUnloggedTables(r.Context(), pool, projectID, schemaName)
-		}
 
 		rows, err := pool.Query(r.Context(),
 			`SELECT id, project_id, action, table_name, column_name, detail, sql_text, created_at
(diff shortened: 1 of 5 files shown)


migrations/000055_ddl_event_trigger.up.sql
```diff
@@ -0,0 +1,266 @@
...
+-- eurobase_migrator, so any role with DDL privileges on a tenant
+-- schema (gateway via SDK DDL, developer pool via console, etc.) can
+-- write through them without needing INSERT on public.schema_changes.
+--
+-- Resilience — every log path is wrapped in EXCEPTION blocks. A bug
+-- in the trigger function MUST NOT cause the user's DDL to roll back.
+
+BEGIN;
+
+-- ── ddl_command_end: CREATE / ALTER ────────────────────────────────
```
STGime, line 36: **Potential blocker:** This migration uses explicit `BEGIN`/`COMMIT`, but golang-migrate (used by this codebase) wraps each `.up.sql` file in its own transaction automatically. Inside that outer transaction:

- `BEGIN` is a no-op (Postgres warns "there is already a transaction in progress")
- `COMMIT` commits the **outer** (golang-migrate) transaction prematurely
- golang-migrate then tries to commit again and gets "WARNING: there is no transaction in progress"

Depending on how golang-migrate handles that, the migration either silently succeeds (fine) or errors out. With `CREATE EVENT TRIGGER`, which **cannot run inside a transaction block** on some Postgres versions, this could be even trickier.

Safest fix: remove the explicit `BEGIN`/`COMMIT` and let golang-migrate handle the transaction. If `CREATE EVENT TRIGGER` requires running outside a transaction, you may need to set `x-migrations-no-transaction` or split the migration.

---

migrations/000055_ddl_event_trigger.up.sql
```diff
@@ -0,0 +1,266 @@
...
+            -- object_identity for table column is 'schema.table.column';
+            -- object_name is the column itself.
+            v_table  := split_part(obj.object_identity, '.', 2);
+            v_column := obj.object_name;
+        ELSIF obj.object_type = 'index' THEN
+            v_action := 'drop_index';
+            v_table  := obj.object_name;
+        ELSE
+            CONTINUE;
+        END IF;
```
STGime: **Bug (low):** For `drop_index`, `obj.object_name` is the **index name**, not the target table. So `schema_changes.table_name` will contain e.g. `idx_users_email` instead of `users`. This is inconsistent with `create_index` above (line ~119) which resolves the actual target table via `pg_index`.

At `sql_drop` time the index is already gone from the catalog, so you can't look up `pg_index.indrelid` — this is a known limitation of sql_drop triggers. Worth either:
- Documenting this in the `detail` JSON (add `"note": "table_name contains index name; target table not resolvable at drop time"`)
- Or renaming the column semantics in the frontend for `drop_index` rows

---

migrations/000055_ddl_event_trigger.up.sql
```diff
@@ -0,0 +1,266 @@
...
+AS $$
+DECLARE
+    cmd          record;
+    v_project_id uuid;
+    v_action     text;
+    v_table      text;
+    v_column     text;
+    v_detail     jsonb;
+    v_query      text := current_query();
+BEGIN
```
STGime, line 58: **Nit (informational):** Event triggers are database-wide — `log_ddl_event()` fires on every DDL statement in the database (platform migrations on `public.*`, River schema migrations, etc.) and then early-exits via the `tenant_*` filter. During a large migration run (e.g. 50+ DDL statements in a deploy), this adds overhead to each statement. Not a blocker, but worth knowing for future performance profiling if deploys slow down.

---

internal/query/ddl_handler.go
```diff
@@ -196,61 +190,17 @@ func HandleSchemaChanges(pool *pgxpool.Pool) http.HandlerFunc {
...
+// logSchemaChange emits an audit log entry for a typed-handler DDL
+// operation. Since #120 the schema_changes insert is handled by the
+// Postgres event triggers (migration 000055), so we no longer write to
+// schema_changes here — that would produce duplicate rows for every
+// typed-handler DDL. The audit_log write stays so DDL actions appear
+// alongside other admin actions in the audit feed.
+//
+// pool is kept in the signature to avoid touching every call site even
+// though the function no longer uses it; remove on the next pass.
 func logSchemaChange(pool *pgxpool.Pool, r *http.Request, projectID, action, tableName string, columnName *string, detail any) {
```
STGime, line 202: Clean removal. The backfill was the root of the #119 bug and no longer has a reason to exist with the event triggers in place. Good that the `pool` parameter is kept to avoid touching every call site — pragmatic.

---

STGime (commented):
Strong PR — event triggers are the right solution here. The trigger functions are well-structured: SECURITY DEFINER for privilege isolation, EXCEPTION blocks to never abort user DDL, tenant-only filtering, and good detail enrichment from `pg_attribute`. The backfill removal and `logSchemaChange` de-duplication are clean.

4 inline comments:

1. **`BEGIN`/`COMMIT` may conflict with golang-migrate (potential blocker)** — golang-migrate wraps each `.up.sql` in its own transaction. The explicit `COMMIT` here would commit the runner's outer transaction prematurely. Additionally, `CREATE EVENT TRIGGER` may not be allowed inside a transaction block on some Postgres versions. Remove the explicit `BEGIN`/`COMMIT` or use the no-transaction migration flag.

2. **`drop_index` stores the index name as `table_name` (low bug)** — Inconsistent with `create_index` which resolves the target table via `pg_index`. At sql_drop time the catalog entry is gone, so this is a known pg limitation, but worth documenting.

3. **Event triggers are database-wide (informational)** — Every DDL statement fires these triggers, including platform migrations. The `tenant_*` filter exits early, but it's overhead on every deploy step.

4. **Backfill removal + `_ = pool`** — Clean and pragmatic.

Fix item 1 before merge; the rest are low-severity.

---

PR #123, opened 2026-05-13
Migration History: audit DDL from the SQL runners (closes #120)

### LINKED ISSUE #120: Migration History: track DDL from all sources (SQL runner, MCP, direct), not just the platform DDL hook

STGime:
Option B from issue #120, after option A (Postgres event triggers) bounced off Scaleway managed Postgres's no-superuser policy (#121 → reverted in #122).

## Approach

Adds a SQL-text detector that scans every successful runner call for DDL shapes and writes one \`schema_changes\` row per detected operation. Wired into all three runner entry points so the gateway-mediated paths — **console SQL editor, MCP \`runSQL\`, SDK \`runSQL\`** — all flow into Migration History.

## What the detector recognises

- \`CREATE TABLE [IF NOT EXISTS] [schema.]name\`
- \`DROP TABLE [IF EXISTS] [schema.]name\`
- \`ALTER TABLE … ADD COLUMN [IF NOT EXISTS] name type\` → \`add_column\`
- \`ALTER TABLE … DROP COLUMN name\`
- \`ALTER TABLE … RENAME TO new_name\` → \`rename_table\`
- \`ALTER TABLE … RENAME COLUMN old TO new\` → \`alter_column\` (\`rename_to\` in detail)
- \`ALTER TABLE … ALTER COLUMN name {TYPE … | SET/DROP NOT NULL | SET/DROP DEFAULT}\`
- \`CREATE [UNIQUE] INDEX [name] ON table\`
- \`DROP INDEX name\` (target table unresolvable — same caveat as the event-trigger \`sql_drop\` path)

False-positive guards: single-quoted strings, line comments, block comments (nest-aware), and dollar-quoted blocks are stripped before matching. \`SELECT 'CREATE TABLE …'\` and \`CREATE FUNCTION\` bodies containing DDL text don't trigger audit rows.

## Wired into

- \`HandleSQL\` (\`/v1/db/sql\` — SDK). Read-only path skips detection because \`ValidateSelectOnly\` rejects DDL upstream; the write path (e.g. \`eb.db.sql()\` via secret key) audits.
- \`HandlePlatformSQL\` (\`/platform/projects/{id}/data/sql\` — console SQL editor + MCP \`runSQL\`).
- \`HandlePlatformSQLTransaction\` (\`/platform/projects/{id}/data/sql/transaction\` — per-statement loop).

## Documented in the console

New **Migration history** subsection in the Database docs page (Database → Migration History timeline). Lists what's tracked, the audited paths, and a yellow callout explaining the limitation that direct DB access (psql / external ORMs / Scaleway DB console) bypasses the audit — plus *why* (event-trigger superuser requirement) and which paths to prefer for a complete audit trail.

## Limitations vs option A

Direct DB access still bypasses the audit. The detector only runs when SQL flows through the gateway. Documented in the issue and the user-facing docs as acceptable given the managed-Postgres privilege ceiling.

## Test plan

- [x] \`go test ./...\` clean (21 table-driven cases for the detector + existing query/gateway suites)
- [ ] Merge + deploy
- [ ] On LiveStylist: \`ALTER TABLE app_users ADD COLUMN test_col text\` via MCP \`runSQL\` → entry appears in Migration History with the correct timestamp + \`type: text\` in detail
- [ ] On any project: \`DROP TABLE\` from the SQL editor → \`drop_table\` row appears
- [ ] Docs page renders the new subsection with the yellow callout

### DIFF
--- a/console/src/routes/(app)/docs/+page.svelte
+++ b/console/src/routes/(app)/docs/+page.svelte
@@ -411,6 +411,24 @@ const eb = createClient({'{'}
 					<li><strong>Bulk delete</strong> &mdash; select multiple rows with checkboxes and delete them at once</li>
 				</ul>
 
+				<h3 class="text-lg font-semibold text-gray-900 mt-6">Migration history</h3>
+
+				<p class="text-sm text-gray-700 leading-relaxed">
+					Database &rarr; <strong>Migration History</strong> shows a timeline of schema changes &mdash; tables created, columns added, indexes dropped, and so on. Each entry records the action, target table/column, and timestamp.
+				</p>
+				<p class="text-sm text-gray-700 leading-relaxed mt-2">
+					The platform tracks DDL from any path that goes through the gateway: the console (Table Editor, SQL Runner), the SDK (<code class="rounded bg-gray-100 px-1.5 py-0.5 text-xs font-mono text-gray-700">eb.db.sql()</code>, <code class="rounded bg-gray-100 px-1.5 py-0.5 text-xs font-mono text-gray-700">eb.schema.createTable()</code>), and MCP tools (<code class="rounded bg-gray-100 px-1.5 py-0.5 text-xs font-mono text-gray-700">runSQL</code>, <code class="rounded bg-gray-100 px-1.5 py-0.5 text-xs font-mono text-gray-700">createTable</code>).
+				</p>
+				<div class="mt-3 rounded-lg border border-amber-200 bg-amber-50 px-4 py-3">
+					<p class="text-sm font-medium text-amber-900">Limitation: direct database access is not audited</p>
+					<p class="mt-1 text-xs text-amber-800 leading-relaxed">
+						Migration History tracks DDL that reaches the platform via the gateway. Schema changes made through a <strong>direct database connection</strong> &mdash; for example, <code class="rounded bg-amber-100 px-1 py-0.5 text-[11px] font-mono">psql</code> connected to your <code class="rounded bg-amber-100 px-1 py-0.5 text-[11px] font-mono">DATABASE_URL</code>, an ORM/migration tool running on your own infrastructure, or the Scaleway database console &mdash; <strong>do not appear</strong> in the history. They are real changes against your schema and your code/queries should still work, but the audit trail won't show them.
+					</p>
+					<p class="mt-2 text-xs text-amber-800 leading-relaxed">
+						This is a deliberate trade-off: capturing every change uniformly would require Postgres superuser privileges, which managed-Postgres providers (Scaleway, RDS, etc.) reserve for themselves. If you need a complete audit trail, prefer the gateway-mediated paths (SQL Runner / SDK / MCP) over direct DB access for any schema-changing operation.
+					</p>
+				</div>
+
 				<h3 class="text-lg font-semibold text-gray-900 mt-6">Using the SDK</h3>
 
 				<div class="relative rounded-lg bg-gray-900 p-4 text-xs font-mono text-green-400 overflow-x-auto">
--- a/internal/query/ddl_detect.go
+++ b/internal/query/ddl_detect.go
@@ -1,0 +1,330 @@
+package query
+
+import (
+	"regexp"
+	"strings"
+)
+
+// DetectedDDL describes a DDL operation that a SQL-runner caller
+// issued. Used by the SQL handlers (HandleSQL / HandlePlatformSQL /
+// HandlePlatformSQLTransaction) to feed the migration-history audit
+// for any DDL that lands outside the typed handlers — i.e. console
+// SQL editor, MCP runSQL, SDK runSQL. Closes #120.
+//
+// Detail mirrors the shape `logSchemaChange` writes: a small JSON
+// object that the migration-history UI can render. Source = "sql"
+// flags these as coming from the SQL-runner path so they can be
+// distinguished from typed-handler audits when needed.
+type DetectedDDL struct {
+	Action     string
+	TableName  string
+	ColumnName *string
+	Detail     map[string]any
+}
+
+// DetectDDL scans a single SQL statement and returns the migration-
+// history-shaped audit rows it implies. Returns an empty slice for
+// statements that don't include any recognised DDL (SELECT / INSERT
+// / UPDATE / DELETE / GRANT / CREATE FUNCTION / CREATE POLICY / etc.).
+//
+// String and dollar-quoted literals, line comments, and block
+// comments are stripped before matching so a SELECT against a
+// table called "create_table" or a CREATE FUNCTION body that
+// happens to contain the text "DROP TABLE" doesn't trigger a
+// false positive.
+//
+// The detector is intentionally lenient — when in doubt it returns
+// nothing rather than mislabel an event. Direct DB access (psql
+// shell, etc.) still bypasses this audit; that's the accepted
+// limitation of the option-B approach (see issue #120).
+func DetectDDL(sql string) []DetectedDDL {
+	clean := stripStringsAndCommentsForDetect(sql)
+	clean = strings.TrimSpace(clean)
+	if clean == "" {
+		return nil
+	}
+
+	var ops []DetectedDDL
+
+	for _, re := range detectors {
+		matches := re.pattern.FindAllStringSubmatch(clean, -1)
+		for _, m := range matches {
+			op := re.build(m)
+			if op != nil {
+				ops = append(ops, *op)
+			}
+		}
+	}
+
+	return ops
+}
+
+// detector pairs a precompiled regex with a function that builds the
+// audit row from the regex's capture groups. Keeping the two
+// together makes adding a new DDL shape a single-block change.
+type detector struct {
+	pattern *regexp.Regexp
+	build   func([]string) *DetectedDDL
+}
+
+// ident matches a bare or double-quoted SQL identifier.
+const ident = `(?:"[^"]+"|[A-Za-z_][A-Za-z0-9_]*)`
+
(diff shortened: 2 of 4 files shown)


internal/query/ddl_detect.go
```diff
@@ -0,0 +1,330 @@
...
+				Action:    "drop_table",
+				TableName: unquoteIdent(m[1]),
+				Detail:    map[string]any{"source": "sql"},
+			}
+		},
+	},
+	{
+		// ALTER TABLE [IF EXISTS] [ONLY] [schema.]name ADD COLUMN [IF NOT EXISTS] name type
+		pattern: regexp.MustCompile(`(?is)\bALTER\s+TABLE\s+(?:IF\s+EXISTS\s+)?(?:ONLY\s+)?` + optSchema + `(` + ident + `)\s+ADD\s+COLUMN\s+(?:IF\s+NOT\s+EXISTS\s+)?(` + ident + `)\s+([A-Za-z][A-Za-z0-9_ ]*)`),
+		build: func(m []string) *DetectedDDL {
```
STGime, line 114: **Limitation (low, acceptable for v1):** Multi-subcommand ALTER TABLE won't detect the second clause. E.g.:

```sql
ALTER TABLE users ADD COLUMN a text, ADD COLUMN b int;
```

The regex requires `\bALTER\s+TABLE` at the start, so only the first `ADD COLUMN a text` matches — the `, ADD COLUMN b int` part lacks the `ALTER TABLE` prefix and is missed. Each regex uses `FindAllStringSubmatch`, but the second ADD COLUMN isn't a standalone match.

Not a blocker — multi-clause ALTER is uncommon in console/MCP usage, and the first column is still captured. Worth a comment or test case documenting the gap so a future contributor doesn't wonder why.

---

internal/query/ddl_detect.go
```diff
@@ -0,0 +1,330 @@
...
+	s := sql
+	i := 0
+	n := len(s)
+	for i < n {
+		c := s[i]
+		switch {
+		case c == '\'':
+			// Single-quoted string. Postgres allows '' as an escape.
+			j := i + 1
+			for j < n {
```
STGime, line 262: The stripper is thorough — nest-aware block comments, `''` escape handling in single-quoted strings, and dollar-quoted bodies all covered. The space-padding to preserve regex offsets is a nice touch that avoids the "two-pass with offset maps" complexity.

One edge case worth considering for a future test: **E-string syntax** `E'...'` or `e'...'` — Postgres supports C-style backslash escapes inside E-strings (`E'it\'s'`). The current scanner handles `''` but not `\'`. This is very rare in practice (especially from a SQL runner UI) so not a blocker.

---

internal/query/sql_handler.go
```diff
@@ -78,6 +108,13 @@ func HandlePlatformSQLTransaction(engine *QueryEngine) http.HandlerFunc {
...
 		}
 
+		// Closes #120 for the multi-statement path. Each statement
+		// committed successfully (the whole tx did), so audit any DDL
+		// shapes we can recognise.
+		for _, stmt := range req.Statements {
+			auditDDLFromSQL(r, engine, stmt)
+		}
+
 		jsonResponse(w, SQLTransactionResponse{
```
STGime, line 118: Good placement — auditing after the successful transaction path means we only log DDL that actually committed. If the tx rolls back, the early-return error path is taken and we never reach here.

---

console/src/routes/(app)/docs/+page.svelte
```diff
@@ -411,6 +411,24 @@ const eb = createClient({'{'}
...
+						Migration History tracks DDL that reaches the platform via the gateway. Schema changes made through a <strong>direct database connection</strong> &mdash; for example, <code class="rounded bg-amber-100 px-1 py-0.5 text-[11px] font-mono">psql</code> connected to your <code class="rounded bg-amber-100 px-1 py-0.5 text-[11px] font-mono">DATABASE_URL</code>, an ORM/migration tool running on your own infrastructure, or the Scaleway database console &mdash; <strong>do not appear</strong> in the history. They are real changes against your schema and your code/queries should still work, but the audit trail won't show them.
+					</p>
+					<p class="mt-2 text-xs text-amber-800 leading-relaxed">
+						This is a deliberate trade-off: capturing every change uniformly would require Postgres superuser privileges, which managed-Postgres providers (Scaleway, RDS, etc.) reserve for themselves. If you need a complete audit trail, prefer the gateway-mediated paths (SQL Runner / SDK / MCP) over direct DB access for any schema-changing operation.
+					</p>
+				</div>
+
 				<h3 class="text-lg font-semibold text-gray-900 mt-6">Using the SDK</h3>
```
STGime, line 432: Clear, honest documentation. The callout explaining *why* (managed Postgres won't grant superuser for event triggers) is exactly the right level of transparency — users understand the constraint isn't a design shortcut.

---

STGime (commented):
Pragmatic option-B after the event trigger approach bounced off Scaleway's no-superuser policy. The regex-based detector is well-engineered: thorough string/comment stripping (nest-aware block comments, dollar-quoting, `''` escapes), 21 table-driven test cases including false-positive guards, and clean wiring into all three SQL runner entry points.

4 inline comments:

1. **Multi-subcommand ALTER TABLE gap (low)** — `ALTER TABLE t ADD COLUMN a text, ADD COLUMN b int` only detects the first column. The regex requires `\bALTER\s+TABLE` as a prefix, so the comma-separated second clause is missed. Acceptable for v1 (uncommon in console/MCP usage), but worth a comment or test case documenting the gap.

2. **E-string syntax edge case (informational)** — `E'it\'s'` uses backslash escapes that the stripper doesn't handle (it only handles `''`). Very rare from a SQL runner UI, not a blocker.

3. **Transaction audit placement** — Correctly placed after the successful commit path so rolled-back DDL is never audited.

4. **Docs callout** — Honest, clear documentation of the limitation and the *why* behind it.

No blockers — the test coverage is solid, the false-positive guards work, and the `logSchemaChange` reuse means the audit rows are consistent with typed-handler DDL. LGTM.

---

Issue #103, opened 2026-05-11
DSAR CSV export: JSONB columns serialised via fmt.Sprintf("%v") — doesn't round-trip


STGime:
## Problem

In \`writeCSV\`:

\`\`\`go
for _, row := range rows {
    record := make([]string, len(cols))
    for i, col := range cols {
        record[i] = fmt.Sprintf(\"%v\", row[col])
    }
    _ = cw.Write(record)
}
\`\`\`

For \`text\`, \`int\`, \`timestamptz\` columns this is fine. For \`jsonb\` (\`metadata\`, \`identity_data\`, \`auth_config\`, etc.) and arrays, Go's \`%v\` produces map[key:value key2:value2] formatting that can't be parsed back as JSON or CSV-imported cleanly. Users receiving a CSV export of complex types get unreadable nested fields.

## Fix

Inside the per-cell loop, type-switch on \`row[col]\`:

\`\`\`go
switch v := row[col].(type) {
case nil:
    record[i] = \"\"
case map[string]interface{}, []interface{}:
    b, _ := json.Marshal(v)
    record[i] = string(b)
case []byte:
    record[i] = string(v)
case time.Time:
    record[i] = v.Format(time.RFC3339)
default:
    record[i] = fmt.Sprint(v)
}
\`\`\`

JSONB columns then serialise as compact JSON strings inside the CSV cell, which any CSV reader handles correctly and which can be re-parsed.

## Files
- \`internal/compliance/export.go\` — \`writeCSV\`

## Notes
Caught during the security audit of #93. Pure UX, not security. Worth fixing before any tenant downloads a CSV export and discovers their data looks like Go map syntax.

STGime:
Closing — fix shipped in PR #125; regression tests TestFormatCSVCell_JSONBRoundTrips + TestFormatCSVCell_Scalars added in PR #145 to lock in jsonb/array roundtrips through json.Marshal (no more Go fmt %v leakage).

Issue #102, opened 2026-05-11
DSAR user export: accepts any UUID for body.UserID without verifying it exists


STGime:
## Problem

\`HandleRequestUserExport\` accepts any string in \`body.UserID\`:

\`\`\`go
var body struct {
    UserID string \`json:\"user_id\"\`
    Format string \`json:\"format\"\`
}
if err := json.NewDecoder(r.Body).Decode(&body); err != nil || body.UserID == \"\" {
    writeExportError(w, \"user_id is required\", http.StatusBadRequest)
    return
}
\`\`\`

If the UUID doesn't match any user in the tenant, the worker still runs: \`DiscoverUserTables\` finds the columns, \`queryTable\` runs with \`WHERE user_id = '<bad-uuid>'\`, every query returns zero rows, the zip ends up containing only \`_metadata.json\`. Compute is wasted; the admin gets a useless 200 KB zip.

Also: a typo in the UUID could let an admin export against a user UUID that exists in another tenant by accident — though the queries are schema-scoped so no actual data leaks, just an empty-but-still-billable export.

## Fix

Before enqueueing, look up \`tenant_<schema>.users WHERE id = \$1\` and return 404 if not present.

## Files
- \`internal/compliance/export_handler.go\` — \`HandleRequestUserExport\` (verify) + new helper in \`internal/compliance/export.go\` for the lookup

## Notes
Caught during the security audit of #93. Not a security finding; the auth gate (\`admin\` of project) is correct. Minor robustness + spend protection.

STGime:
Closing — fix shipped in PR #125; regression test TestUserExistsInTenant_TrueFalse added in PR #145 to lock in the behaviour (true for present users, false for random UUIDs).

Issue #100, opened 2026-05-11
DSAR export: no audit_log write on export request / completion


STGime:
## Problem

\`HandleRequestTenantExport\` / \`HandleRequestUserExport\` / \`HandleSelfServeExport\` don't write to \`public.audit_log\` even though \`auditSvc\` is already injected and the rest of the platform uses it.

A malicious admin who exfiltrates project data via the DSAR endpoint leaves only the row in \`export_requests\` as evidence. The presigned URLs expire after 1h, the files after 7 days, and an admin could (in principle) suppress the row by manipulating the underlying DB if they have other access paths.

## Fix

Emit two audit events per export lifecycle:

- \`compliance.export.requested\` — when CreateExportRequest succeeds (project_id, target user_id if any, format, requested_by)
- \`compliance.export.completed\` — when MarkCompleted runs in the worker (export_id, s3_key, file_size)

Self-serve path should also emit \`compliance.export.self_requested\` so end-user exports are logged.

## Files
- \`internal/compliance/export_handler.go\` — three handler call sites
- \`internal/compliance/export.go\` — \`MarkCompleted\` (or call from the worker after MarkCompleted)
- \`internal/workers/export.go\` — TenantExportWorker / UserExportWorker

## Notes
Caught during the security audit of #93. Defence-in-depth for compliance reporting; not a strict GDPR requirement but a useful evidence trail.

STGime:
Failed-paths follow-up closed by PR #145. compliance.export.failed now emitted from all 3 handler enqueue-failure paths and 3 worker failure stages (resolve_project, build_zip, upload), with stage + error metadata. The previously-defined-but-unused ActionExportFailed constant is now wired up.

PR #125, opened 2026-05-15
Quick wins: CI Node 24 + DSAR audit + UUID check + CSV JSONB (closes #96 #100 #102 #103)

### LINKED ISSUE #96: CI: GitHub Actions warning — Node.js 20 deprecation on actions/checkout, setup-go, setup-node
## Warning

Every CI/CD run currently emits the same deprecation warning across all jobs (test-go, test-console, test-mcp, test-functions-runner):

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: \`actions/checkout@v4\`, \`actions/setup-go@v5\`, \`actions/setup-node@v4\`. Actions will be forced to run with Node.js 24 by default starting **June 2nd, 2026**. Node.js 20 will be removed from the runner on **September 16th, 2026**.

Reference: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Most recent run: https://github.com/STGime/euroback/actions/runs/25626103070

## Impact
- Today: warning only, no functional break.
- 2026-06-02: actions auto-upgraded to Node 24 — could surface bugs if any of the action versions we pin haven't been validated against Node 24 by their maintainers.
- 2026-09-16: hard cutover, anything still on Node 20 breaks.

## Fix options

1. **Wait for new major versions** (\`actions/checkout@v5\`, \`actions/setup-go@v6\`, \`actions/setup-node@v5\`) and bump the pins. Cleanest path; we just need to track when each releases.
2. **Opt in early** by setting \`FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true\` either as a workflow env or on the runner. Validates compatibility now.

Recommended: do (2) first as a single-line workflow env to make sure nothing breaks under Node 24, then bump action versions when each maintainer ships.

## Touch points
- \`.github/workflows/ci.yml\` — every job uses these three actions.

Five-minute change once decided.

STGime:
Four small fixes off the backlog, bundled because each is too short to PR alone.

## #96 — CI: opt in to Node 24 now

\`actions/checkout@v4\`, \`setup-go@v5\`, \`setup-node@v4\` are pinned on Node 20 which GitHub auto-upgrades to Node 24 on **2026-06-02** and removes outright on **2026-09-16**. One-line workflow env (\`FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true\`) flips CI to Node 24 now so any incompat surfaces on a normal run, not the cutover day. The env becomes a no-op once each action ships a Node-24-native major; we can drop it then.

## #100 — DSAR export: audit_log entries on request + completion

A malicious admin who exfiltrates project data via DSAR used to leave **only** a row in \`export_requests\` as evidence — deletable with DB access. Now every export emits two audit rows: one at request time (handler) and one at completion (worker after \`MarkCompleted\`).

Four new constants in \`internal/audit\`:
- \`compliance.export.requested\`
- \`compliance.export.self_requested\` (end-user-initiated)
- \`compliance.export.completed\`
- \`compliance.export.failed\` (reserved for failure-path follow-up)

Wired into \`HandleRequestTenantExport\`, \`HandleRequestUserExport\`, \`HandleSelfServeExport\`, plus \`TenantExportWorker\` and \`UserExportWorker\`.

## #102 — DSAR user export: verify user exists before enqueueing

\`HandleRequestUserExport\` used to accept any UUID. If it didn't match a tenant user, the worker ran zero-row queries to completion and produced an almost-empty zip — useless to the recipient, billable compute, and the admin's \"export ran fine\" message was indistinguishable from success.

New \`ExportService.UserExistsInTenant\` does a one-round-trip \`SELECT EXISTS\` against \`<schema>.users\`. Runs **before** the rate limit so a typo doesn't consume the daily slot. 404 if the user isn't there.

## #103 — DSAR CSV: JSONB / array columns round-trip via json.Marshal

\`writeCSV\` previously rendered every cell via \`fmt.Sprintf(\"%v\", v)\`. For jsonb (\`metadata\`, \`identity_data\`, \`auth_config\`) and arrays, that produces Go's \`map[k:v]\` / \`[a b c]\` form, which is neither valid JSON nor CSV-importable. Recipients couldn't actually use the export.

New \`formatCSVCell\` type-switch:

| Type | Format |
|---|---|
| \`nil\` | \`\"\"\" |
| \`string\` | as-is |
| \`time.Time\` | RFC3339Nano |
| \`[]byte\` | \`string(v)\` |
| \`map\` / \`slice\` (jsonb, array) | \`json.Marshal\` |
| default | \`fmt.Sprint\` |

GDPR's \"machine-readable structured format\" obligation now actually produces parseable output.

## Test plan

- [x] \`go test ./...\` clean
- [ ] Merge + deploy
- [ ] Trigger a tenant DSAR export → \`compliance.export.requested\` + \`compliance.export.completed\` rows appear in Compliance → Audit Log
- [ ] Trigger a user DSAR export with a non-existent UUID → 404
- [ ] Open a previous DSAR export CSV with jsonb columns → fields parse as JSON
- [ ] Next CI run shows no Node 20 deprecation warnings

### DIFF
--- a/internal/compliance/export.go
+++ b/internal/compliance/export.go
@@ -99,6 +99,33 @@ func (s *ExportService) CreateExportRequest(ctx context.Context, projectID strin
 	return &req, nil
 }
 
+// UserExistsInTenant reports whether the given user_id exists in the
+// tenant's `users` table. Closes #102: HandleRequestUserExport used
+// to accept any UUID and enqueue a worker even if the user wasn't in
+// the tenant — producing a useless near-empty zip and burning compute.
+//
+// schema_name is resolved from projects in one round-trip; we use a
+// single QueryRow with EXISTS so the call is cheap (~1ms on hot path).
+// quoteIdent guards the schema name even though it comes from a
+// platform-controlled column, on the principle that defence in depth
+// for identifier interpolation is free.
+func (s *ExportService) UserExistsInTenant(ctx context.Context, projectID, userID string) (bool, error) {
+	var schemaName string
+	err := s.Pool.QueryRow(ctx,
+		`SELECT schema_name FROM projects WHERE id = $1`, projectID,
+	).Scan(&schemaName)
+	if err != nil {
+		return false, fmt.Errorf("resolve project schema: %w", err)
+	}
+
+	var exists bool
+	q := fmt.Sprintf(`SELECT EXISTS (SELECT 1 FROM %s.users WHERE id = $1)`, quoteIdent(schemaName))
+	if err := s.Pool.QueryRow(ctx, q, userID).Scan(&exists); err != nil {
+		return false, fmt.Errorf("check user existence: %w", err)
+	}
+	return exists, nil
+}
+
 // GetExportRequest retrieves an export request by ID, scoped to a project.
 func (s *ExportService) GetExportRequest(ctx context.Context, exportID, projectID string) (*ExportRequest, error) {
 	var req ExportRequest
@@ -592,10 +619,51 @@ func writeCSV(w io.Writer, rows []map[string]interface{}) (int, error) {
 	for _, row := range rows {
 		record := make([]string, len(cols))
 		for i, col := range cols {
-			record[i] = fmt.Sprintf("%v", row[col])
+			record[i] = formatCSVCell(row[col])
 		}
 		_ = cw.Write(record)
 	}
 	cw.Flush()
 	return len(rows), cw.Error()
+}
+
+// formatCSVCell renders one column value for the DSAR CSV export.
+//
+// Closes #103. The previous fmt.Sprintf("%v", …) path mangled jsonb
+// and array columns into Go's map[k:v] / [a b c] string forms, which
+// is neither valid JSON nor parseable back into anything useful. The
+// DSAR recipient needed those columns intact to satisfy GDPR's
+// "machine readable, structured format" obligation.
+//
+// Behaviour by type:
+//   - nil               → empty string (CSV null)
+//   - map / slice / any composite → json.Marshal so jsonb /
+//                         text[] / row arrays round-trip
+//   - time.Time          → RFC3339Nano (matches what pgx emits and
+//                         what most spreadsheet apps recognise)
+//   - []byte             → string conversion (Postgres bytea
+//                         literal; if it happens to be UTF-8 text,
+//                         the cell is human-readable, otherwise
+//                         the recipient gets the raw bytes the
+//                         pg driver returned)
+//   - everything else    → fmt.Sprint, same as before
+func formatCSVCell(v any) string {
+	switch v := v.(type) {
+	case nil:
+		return ""
+	case string:
+		return v
+	case time.Time:
+		return v.Format(time.RFC3339Nano)
+	case []byte:
+		return string(v)
+	case map[string]interface{}, []interface{}:
+		b, err := json.Marshal(v)
+		if err != nil {
+			return fmt.Sprintf("%v", v)
+		}
+		return string(b)
+	default:
+		return fmt.Sprint(v)
+	}
 }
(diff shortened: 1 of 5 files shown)


internal/compliance/export_handler.go
```diff
@@ -104,6 +119,22 @@ func HandleRequestUserExport(exportSvc *ExportService) http.HandlerFunc {
...
+			return
+		}
+		if !exists {
+			writeExportError(w, "user not found", http.StatusNotFound)
+			return
+		}
+
 		limited, err := exportSvc.CheckRateLimit(r.Context(), projectID, &body.UserID)
 		if err != nil {
 			writeExportError(w, "internal error", http.StatusInternalServerError)
```
STGime, line 140: Good placement — before the rate limit so a typo'd UUID doesn't burn the 24-hour slot. The `quoteIdent(schemaName)` is the right call even though `schema_name` is platform-controlled — defense in depth for identifier interpolation is essentially free.

---

internal/compliance/export.go
```diff
@@ -592,10 +619,51 @@ func writeCSV(w io.Writer, rows []map[string]interface{}) (int, error) {
...
+		return string(v)
+	case map[string]interface{}, []interface{}:
+		b, err := json.Marshal(v)
+		if err != nil {
+			return fmt.Sprintf("%v", v)
+		}
+		return string(b)
+	default:
+		return fmt.Sprint(v)
+	}
```
STGime, line 668: Good type switch. One thing to be aware of: pgx can also return jsonb as `map[string]any` (alias for `map[string]interface{}`) or Postgres arrays as typed Go slices (e.g. `[]string`, `[]int64`) depending on the column's OID and pgx version. These typed slices would fall through to the `default: fmt.Sprint` path and produce the `[a b c]` form.

If you hit this in practice, a quick fix is to add a `reflect`-based fallback before the default:

```go
case reflect.Map, reflect.Slice:
    b, _ := json.Marshal(v)
    return string(b)
```

Not a blocker — `rows.Values()` returns `[]interface{}` for arrays in the current pgx version, so the explicit case handles the hot path.

---

internal/workers/export.go
```diff
@@ -56,6 +56,24 @@ func (w *TenantExportWorker) Work(ctx context.Context, job *river.Job[jobs.Tenan
...
+				"s3_key":    s3Key,
+				"file_size": buf.Len(),
+				"rows":      totalRows,
+				"scope":     "tenant",
+			}))
+	}
+
 	logger.Info("tenant export completed", "size", buf.Len(), "rows", totalRows)
 	return nil
 }
```
STGime, line 79: Clean: empty actor (`"", ""`) on the worker side is correct since the request-time audit already records who initiated it. The two rows link via `target_id = export_id`. The `s3_key` + `file_size` + `rows` metadata on completion gives a forensic investigator everything they need to assess what was extracted.

---

STGime (commented):
Four clean small fixes, well-scoped for a bundle PR. Each one stands on its own:

- **#96 (CI Node 24)** — Proactive. One env var, surfaces incompat before the cutover day, becomes a no-op later.
- **#100 (DSAR audit)** — Defence-in-depth: request-time + completion-time audit rows linked by export_id. Covers all three entry points (tenant, user, self-serve) plus both workers. The `ActionExportFailed` constant reserved for follow-up is tidy.
- **#102 (UUID check)** — `UserExistsInTenant` before rate limit is the right ordering. Prevents wasted compute and the false "success" UX on non-existent user exports.
- **#103 (CSV JSONB)** — `formatCSVCell` type switch fixes the GDPR "machine-readable format" obligation. `json.Marshal` for maps/slices is the correct path.

3 inline comments, all low-severity. No blockers — LGTM.

---

PR #136, opened 2026-05-16
Cost: in-cluster Redis + replicas=1 during beta (~€60-90/mo saved)

STGime:
April Scaleway bill was ~€97; May was tracking ~€127. Two changes cut the recurring cost to ~€45/mo with no functional regression at our current beta scale.

| Change | Saving |
|---|---|
| In-cluster Redis (replaces Managed RED1-micro) | €34-59/mo |
| Kapsule pool 2 → 1 nodes + replicas=1 across the public-path deployments | ~€28/mo (after the operator follow-up) |
| **Total** | **~€60-90/mo** |

### \`deploy/k8s/redis.yaml\` — in-cluster Redis StatefulSet + Service

Single-replica \`redis:7-alpine\` with \`--save \"\" --appendonly no\` (pure in-memory), 192 MB max with \`allkeys-lru\` eviction, 25m/64Mi requests / 200m/256Mi limits. Headless Service so the gateway resolves \`redis.eurobase.svc.cluster.local\`.

The existing consumers in the codebase already log \`REDIS_URL not set, X disabled\` graceful fallbacks (\`cmd/gateway/main.go:124\` for rate limiting, \`:210\` for realtime cross-instance fan-out), so the failure mode is well-understood.

### \`deploy/k8s/{gateway,mcp-server,functions}.yaml\` — replicas: 2 → 1

Total pod requests across the cluster sum to ~1.1 CPU / 1.3 GB — fits on one DEV1-M (3 vCPU / 4 GB) with > 50% headroom. The HPA on functions follows (\`minReplicas: 2 → 1\`, max stays at 10).

**Trade-off:** no redundancy. A node reboot = ~3 min outage. Consistent with the \"beta test system, no uptime guarantees\" disclaimer in the invitation email.

### \`.github/workflows/ci.yml\` — wire Redis into the deploy

- Apply \`redis.yaml\` before the gateway rollout so the new URL resolves on first start.
- Add \`kubectl rollout status statefulset/redis\` to the verify step.

### \`docs/runbooks/cost-tier.md\` — beta vs production cost-tier

Documents both directions:
- **Beta → production**: when the first paying Pro customer signs up (tracked in #135), three reversible steps in fixed order: pool autoscaler → replicas → managed Redis.
- **Production → beta**: today's PR, written down so we can repeat it on a quiet weekend.

## Operator follow-up (after merge — not automated)

The PR ships the manifests + CI wiring. The actual cutover requires two operator commands:

[...]

Issue #135, opened 2026-05-16
Scale infra back up to production tier when first paying Pro customer signs up


STGime:
**Context:** We just cut the Scaleway bill from ~€127/mo to ~€45/mo by going to beta-tier infra: in-cluster Redis (vs Managed RED1-micro), single Kapsule node (vs 2× DEV1-M), and replicas=1 on the gateway / mcp-server / functions. See PR (cost: in-cluster Redis + replicas=1).

That trade-off is fine for the beta — the invitation email explicitly disclaims uptime guarantees — but **a paying Pro customer expects redundancy**. Production-tier infra costs ~€127/mo and tolerates a Kapsule node restart without downtime.

## When to do this

**First paying Pro customer signs up** (Mollie billing → \`compliance.export.requested\` is the first real revenue signal, but most likely it's just \"first time we charge €19/mo\"). That's the moment to flip back.

Don't wait for outage complaints — the moment cash starts flowing, the contract changes.

## How to do this

Follow [\`docs/runbooks/cost-tier.md\`](../tree/main/docs/runbooks/cost-tier.md), section **\"Beta → production\"**. Three steps in fixed order:

1. **Kapsule pool autoscaler back to min=1, max=2** (~€28/mo) — buys back redundancy on the gateway path.
2. **Restore replicas=2 on gateway, mcp-server, functions** (no extra cost — the new node has room). Update the manifests so the next deploy doesn't regress.
3. **Optional: swap in-cluster Redis for managed RED1-micro** (~€34-59/mo) — only when cross-AZ realtime or durable rate-limit counters matter.

Each step is independently reversible. The runbook has the exact \`scw\` / \`kubectl\` commands.

## Acceptance

- [ ] Kapsule pool is at 2 nodes
- [ ] gateway/mcp-server/functions Deployments back at replicas=2 (manifests too)
- [ ] Both nodes consistently scheduled (verify with \`kubectl get pods -o wide\`)
- [ ] Optionally: managed Redis live, in-cluster Redis StatefulSet deleted
- [ ] Scaleway cost alert updated from €50/mo to €150/mo

## Don't merge this issue

This is intentionally a checklist for a future calendar event, not code work. Close when the steps above are done.

PR #137, opened 2026-05-16
Add scripts/cost-tier-to-beta.sh — automate the post-#136 cutover

STGime:
Companion to #136 / \`docs/runbooks/cost-tier.md\`. Single script that runs the three operator steps to flip production from \"€127/mo HA tier\" to \"€45/mo beta tier\", with verification between every step and a typed-confirmation gate on the only destructive action.

## Why

Three manual commands were left for the operator after #136 merged:

1. patch \`REDIS_URL\` in \`eurobase-secrets\` to point at the in-cluster service
2. shrink the Kapsule pool to 1 node
3. delete the managed Scaleway Redis cluster (RED1-micro)

The manual sequence hit two macOS Terminal / shell footguns on first attempt (multi-line paste joining, \`VAR=val command\` scoping) — silently wrote an empty \`REDIS_URL\` and dropped both rate limiting and realtime fan-out. Script avoids that by using a pure pipeline (\`printf | base64 | tr -d '\\n' | xargs\`).

## What it does

| Phase | Action | Verification |
|---|---|---|
| Pre-flight | StatefulSet \`redis/eurobase\` is Ready 1/1 | Pings via one-shot pod, expects PONG |
| Step 1 | Patch Secret → restart gateway + worker | Greps gateway logs for the literal \`rate limiter redis connected\` and \`realtime redis bridge connected\` lines from \`internal/ratelimit/limiter.go:71\` and \`internal/realtime/redis.go:39\`. Bails if neither appears. |
| Step 2 | \`scw k8s pool update … min-size=1 max-size=1 size=1\` | Polls \`kubectl get nodes\` for up to 6 min, bails if it doesn't converge to 1 |
| Step 3 | Delete managed Redis (\`scw redis cluster delete\`) | Typed confirmation: must enter exact string \`delete managed redis\`. Skipping is safe, prints the manual command and exits. |

All resource IDs hardcoded (pool ID, managed Redis cluster ID) — verified 2026-05-16. They don't change unless we reprovision.

## Verified in production today

Ran the script (with one tiny fix mid-stream — \`scw\` uses top-level \`min-size\` / \`max-size\` args, not under an \`autoscaler.\` prefix as I'd first written). Output confirmed all three phases succeeded:

- Gateway logs: \`rate limiter redis connected addr=redis.eurobase.svc.cluster.local:6379\`
- \`kubectl get nodes\` converged from 2 → 1 within ~80 seconds
- Managed RED1-micro cluster deletion accepted by the Scaleway API

The committed version of the script contains the corrected \`scw\` syntax.

## Cross-referenced

- \`docs/runbooks/cost-tier.md\` now mentions the one-shot \`bash deploy/scripts/cost-tier-to-beta.sh\` above the manual sequence.
- The inverse direction (\"Beta → production\" in the runbook) stays manual for now; #135 tracks when to use it.

### DIFF
--- a/deploy/scripts/cost-tier-to-beta.sh
+++ b/deploy/scripts/cost-tier-to-beta.sh
@@ -108,7 +108,15 @@ bold "  checking gateway logs for the redis-connect line..."
 # The Go gateway logs structured JSON. Look for either of the two
 # success messages from internal/ratelimit/limiter.go:71 or
 # internal/realtime/redis.go:39.
-sleep 4   # give the new pods a beat to emit the line
+#
+# Wait for at least one gateway pod to be Ready before grepping —
+# a fixed sleep was the original approach but cold starts (image
+# pull, slow node) could blow past 4s and miss the log line, then
+# the script would bail and force a re-run. `kubectl wait` blocks
+# until the pod's Ready condition is true (or 60s, whichever
+# comes first) and exits non-zero on the timeout so we still error
+# out cleanly if the rollout itself is broken.
+kubectl wait --for=condition=Ready pod -l app=gateway -n "$NAMESPACE" --timeout=60s >/dev/null
 log_hit=$(kubectl logs -n "$NAMESPACE" -l app=gateway --tail=400 --all-containers=true 2>&1 \
   | grep -iE 'rate limiter redis connected|realtime redis bridge connected' || true)
 if [[ -z "$log_hit" ]]; then
--- a/deploy/scripts/cost-tier-to-beta.sh
+++ b/deploy/scripts/cost-tier-to-beta.sh
@@ -1,0 +1,183 @@
+#!/usr/bin/env bash
+#
+# cost-tier-to-beta.sh — cut Eurobase's Scaleway bill from ~€127/mo to ~€45/mo.
+#
+# Companion to docs/runbooks/cost-tier.md and PR #136. Runs the three
+# post-merge operator steps in sequence, with verification + a final
+# confirmation prompt before the only destructive action (deleting the
+# managed Redis instance).
+#
+# What this does:
+#   1. Patches REDIS_URL in the eurobase-secrets Secret to point at the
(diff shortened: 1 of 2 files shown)


STGime (commented):
Well-automated companion to #136. The script is production-verified and addresses real shell footguns that bit the manual attempt (multi-line paste joining, `VAR=val command` scoping writing empty `REDIS_URL`).

**Good patterns:**
- Pre-flight checks (StatefulSet ready + one-shot pod PING) catch the "Redis isn't up yet" case before touching secrets.
- Log verification (`grep -iE 'rate limiter redis connected|realtime redis bridge connected'`) after the gateway roll confirms the cutover actually worked — not just that pods restarted.
- Pipeline form (`printf | base64 | tr -d '\n' | xargs`) for the Secret patch avoids the shell-scope footguns from the manual attempt.
- Typed confirmation (`delete managed redis`) before the only irreversible action. Skipping prints the manual command and exits cleanly.
- 6-min drain timeout with progress (`nodes: N`) is good operator UX.

**One note:** The `sleep 4` before log-grepping (line ~128) is a fixed wait that could miss slow-starting pods. On a cold start with image pull, the gateway might take longer than 4s. If the grep misses, the script bails — the operator has to re-run. Not a blocker (the script is idempotent), but a `kubectl wait --for=condition=Ready pod -l app=gateway --timeout=60s` before the log check would be more robust.

**Overlap with #136:** The diff includes #136's changes (redis.yaml, replicas, runbook) because both branches share a base. If #136 is already merged, this should rebase cleanly to just the script + runbook update.

LGTM — ship after rebase on latest main.

---

STGime (commented):
Feedback addressed. The `sleep 4` before log-grepping is replaced with `kubectl wait --for=condition=Ready pod -l app=gateway --timeout=60s` — handles cold starts and slow image pulls cleanly. Diff is now just the script + runbook update (rebased off the #136 overlap). LGTM — no remaining issues.

---

PR #142, opened 2026-05-16
Compliance: add Request Single-User Export UI (page promise → reality)

STGime:
The Data Export tab subtitle has promised \"Export all project data **or a specific user's data**\" since DSAR shipped, but the UI only had a project-export button. The backend route \`POST /platform/projects/{id}/compliance/user-export\` has been live the whole time — only the picker was missing.

## What lands

A new **Request Single-User Export** card beneath the project-export card:

- **Debounced search box** (250ms) — hits the existing \`api.listEndUsers(projectId, { search, limit: 10 })\` endpoint. Matches email / display name / phone.
- **Click-to-select dropdown** — each row shows email + display name + first 8 chars of the UUID; click to pick.
- **Confirmation chip** — selected user shows as a row with the full UUID + a Clear button. Submit button stays disabled until a user is picked, so the wrong UUID can't slip through.
- **Format selector** (JSON / CSV) + **Export User Data** button. Wired to \`api.requestUserExport(projectId, userId, format)\` which was already in the SDK, just unused.
- **Friendly 429 message** — \"Rate limit: 1 user export per 24 hours. Please wait.\" matches the project-export path's treatment.

## Export History — Type column

Now shows the first 8 chars of the user UUID next to \"User\" so multiple per-user exports stay distinguishable at a glance:

\`User a1b2c3d4…\` instead of just \`User\` for every row.

## Why this matters

Closes the gap between the page subtitle and the page itself. Also lifts a key DSAR pitch — \"respond to one of your users' Article 15 request in seconds, with their data extracted across every table they touched\" — from \"technically possible via API\" to \"two clicks in the console.\"

## Test plan

- [ ] Merge + deploy
- [ ] Compliance → Data Export → search for a known user email → row appears in dropdown
- [ ] Pick the user → confirmation chip shows full UUID
- [ ] Click \"Export User Data\" → row appears in Export History with Type = \`User <short>\`
- [ ] Click rapidly twice → 429 with the friendly rate-limit message

### DIFF
--- a/console/src/routes/(app)/p/[id]/compliance/+page.svelte
+++ b/console/src/routes/(app)/p/[id]/compliance/+page.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { page } from '$app/stores';
-	import { api, type DPAReport, type AuditLogEntry } from '$lib/api.js';
+	import { api, type DPAReport, type AuditLogEntry, type EndUser } from '$lib/api.js';
 	import { onMount } from 'svelte';
 
 	let projectId = $derived($page.params.id);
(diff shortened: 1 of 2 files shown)


STGime (commented):
Closes the gap between the page subtitle's promise and reality. The UX is well thought out:

- **Debounced search (250ms)** — won't spam the server on fast typing, but feels instant to the user.
- **Dropdown** — shows email + display_name + short UUID. Three signals so the operator can confidently identify the right user even with similar emails.
- **Confirmation chip** — full UUID visible before submit. Prevents "exported the wrong Alex" mistakes.
- **Button disabled without selection** — can't accidentally fire an empty request.
- **Rate-limit regex** — `if (/rate.?limit/i.test(msg))` gives a friendly message instead of the raw API error. Nice DX.
- **Export history type column** — `User a1b2c3d4…` distinguishes multiple per-user exports at a glance. Good improvement over the bare "User" label.

One small observation: the search input doubles as a "paste UUID" affordance (placeholder text says so), but `api.listEndUsers` with a UUID string as `search` would need the backend to match on `id` in addition to email/name/phone. If the endpoint already does this (likely, since it's a general search), great. If not, pasting a UUID would return no results even though the user exists.

LGTM — no blockers.

---

Issue #101, opened 2026-05-11
DSAR export: rate limit counts failed exports, blocks retry for 24h / 1h


STGime:
## Problem

\`compliance.CheckRateLimit\` counts every row in \`export_requests\` created within the window, regardless of status:

\`\`\`sql
SELECT COUNT(*) FROM export_requests
WHERE project_id = \$1 AND user_id = \$2 AND created_at > now() - interval '24 hours'
\`\`\`

If a user's export fails (worker crash, transient S3 error, OOM per #N), they're locked out from retrying for 24h. Same for the tenant path with the 1h window.

## Fix

Add \`status NOT IN ('failed', 'pending')\` to the WHERE clause so only successful or in-progress exports count toward the limit. A failed request shouldn't block legitimate retry.

## Files
- \`internal/compliance/export.go\` — \`CheckRateLimit\`

## Notes
Small change, ~2 lines. Caught during the security audit of #93.

STGime:
Closed by PR #145. CheckRateLimit now filters status <> 'failed' in both 24h-user and 1h-tenant windows. Failed exports no longer count against the budget. Regression test TestCheckRateLimit_IgnoresFailedExports covers fresh→failed→completed for both scopes.

Issue #99, opened 2026-05-11
DSAR export: BuildTenantExportZip materialises whole archive in RAM (OOM risk)


STGime:
## Problem

\`compliance.BuildTenantExportZip\` and \`BuildUserExportZip\` write the entire zip into a \`bytes.Buffer\`, then upload it to S3 in one \`UploadObject\` call. For a tenant with many tables × the per-table \`LIMIT 100000\` cap, this can balloon to multiple GB resident in the worker pod:

> 50 tables × 100k rows × ~1 KB JSON ≈ 5 GB

That OOMs the worker pod and kills any concurrent jobs on the same instance.

## Fix

Stream the zip directly to S3 via multipart upload, or to a tempfile that's then uploaded once. \`storage.S3Client\` doesn't currently expose a streaming upload method — would need a new one (e.g. \`UploadObjectStreaming(ctx, bucket, key, reader, contentType) error\`) that wraps the multipart API.

Worker side: \`zip.NewWriter\` accepts any \`io.Writer\`, so we can write straight into a \`io.Pipe\` paired with the S3 multipart reader.

## Files
- \`internal/compliance/export.go\` — \`BuildTenantExportZip\`, \`BuildUserExportZip\`
- \`internal/workers/export.go\` — TenantExportWorker / UserExportWorker call sites
- \`internal/storage/s3.go\` — new streaming upload helper

## Notes
Caught during the security audit of #93. Not a security finding (the buffered build doesn't leak data); it's an operational reliability concern that scales with tenant size.

PR #145, opened 2026-05-18
DSAR cluster: stream zip + audit on failure + rate-limit fix (#99, #100, #101)

STGime:
## Summary

- **#99** — `BuildTenantExportZip`/`BuildUserExportZip` (returned `*bytes.Buffer`) replaced with `WriteTenantExport`/`WriteUserExport` (`io.Writer` destination). Workers stream into a temp file, then upload the rewound `*os.File` to S3. Memory is bounded by one row + zip buffer regardless of table size.
- **#100 (failed-paths follow-up)** — Three handler enqueue-failure paths and three worker failure stages (`resolve_project`, `build_zip`, `upload`) now emit `compliance.export.failed` with `stage` + `error` metadata. Makes the SDK README's "rate-limited and audited" claim true for the failure path.
- **#101** — `CheckRateLimit` SQL now filters `status <> 'failed'`. A worker crash no longer locks the requester out for 24h.
- **#102, #103 lock-in tests** — Both were fixed in PR #125 but had no regression test. Added pure unit tests for `formatCSVCell` (jsonb maps, arrays, scalars) and an integration test for `UserExistsInTenant` (true for present users, false for random UUIDs).
- **#3** — Verified live (cert Ready=True until 2026-06-24); `docs/TODO.md` ticked, issue closed separately.

### Streaming detail

`streamQueryToZip` is the new single helper for both functions:

```
rows.Next() → row map → CSV writer / manual-JSON encoder → zip entry → io.Writer
```

JSON-array shape (`[{...},{...}]`) preserved on the wire — no consumer change for the SDK, console download, or any scripts customers wrote against v0.4. Empty queries skip the zip entry entirely (matches previous behaviour).

### Drive-bys

- `sdk/js/test/auth-export.mjs`: `setSession()` schedules a refresh-token `setTimeout` ~59 min out that kept Node alive after assertions passed, so `npm test` hung indefinitely. Pre-existing — not regression of this PR — but it was blocking the test suite from completing. Added explicit `await eb.auth.signOut()` at end.

## Files

| File | Change |
|---|---|
| `internal/compliance/export.go` | Streaming refactor; `CheckRateLimit` SQL fix; old `Build*ExportZip` removed |
| `internal/compliance/export_handler.go` | Audit emission on enqueue failure (3 handlers) |
| `internal/workers/export.go` | Temp-file streaming; `failExport` closure for audit on each stage |
| `internal/compliance/export_test.go` | **NEW** — 521 lines, 6 test functions |
| `docs/TODO.md` | #3 wildcard TLS section ticked |
| `sdk/js/test/auth-export.mjs` | Clean Node exit |

## Test plan

- [x] All 4 SDK test files pass individually (token-propagation, realtime-url, schedules, auth-export)
- [x] `npm test` now completes (was hanging on auth-export.mjs)
- [ ] CI runs full integration suite with `DATABASE_URL` set → `TestCheckRateLimit_IgnoresFailedExports`, `TestUserExistsInTenant_TrueFalse`, `TestWriteTenantExport_StreamsZipStructure`, `TestWriteUserExport_FiltersToOneUser` execute against real Postgres
- [ ] After merge: verify a project DSAR export still works end-to-end on superapp; confirm the audit log shows new entries for any failed export

### DIFF
--- a/internal/compliance/export_test.go
+++ b/internal/compliance/export_test.go
@@ -1,0 +1,521 @@
+package compliance
+
+import (
+	"archive/zip"
+	"bytes"
+	"context"
+	"encoding/csv"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// ── #103: formatCSVCell — pure unit tests (no DB) ────────────────────────────
+
+// TestFormatCSVCell_JSONBRoundTrips locks in #103: a JSONB column
+// pulled out as map[string]interface{} must serialise as parseable JSON
+// in the CSV cell, NOT Go's fmt "%v" form (`map[key:val]`).
+func TestFormatCSVCell_JSONBRoundTrips(t *testing.T) {
+	cases := []struct {
+		name string
+		in   any
+		// We don't compare raw strings because map iteration order is
+		// non-deterministic; instead we re-parse the output as JSON
+		// and verify it's structurally equal.
+		want any
+	}{
+		{
+			name: "jsonb object",
+			in:   map[string]interface{}{"k": "v", "n": float64(7)},
+			want: map[string]interface{}{"k": "v", "n": float64(7)},
+		},
+		{
+			name: "text array",
+			in:   []interface{}{"a", "b", "c"},
+			want: []interface{}{"a", "b", "c"},
+		},
+		{
+			name: "nested",
+			in: map[string]interface{}{
+				"settings": map[string]interface{}{"theme": "dark", "count": float64(2)},
+				"tags":     []interface{}{"x", "y"},
+			},
+			want: map[string]interface{}{
+				"settings": map[string]interface{}{"theme": "dark", "count": float64(2)},
+				"tags":     []interface{}{"x", "y"},
+			},
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			s := formatCSVCell(c.in)
+			if s == "" {
+				t.Fatalf("got empty string for input %v", c.in)
+			}
+			if strings.HasPrefix(s, "map[") {
+				t.Fatalf("regression of #103: cell formatted via Go fmt %%v: %q", s)
+			}
+			var got any
+			if err := json.Unmarshal([]byte(s), &got); err != nil {
+				t.Fatalf("cell not parseable as JSON: %v\nraw: %q", err, s)
+			}
+			gotJSON, _ := json.Marshal(got)
+			wantJSON, _ := json.Marshal(c.want)
+			// Re-marshal both sides to canonicalise.
+			if !bytes.Equal(gotJSON, wantJSON) {
+				t.Errorf("round-trip mismatch:\n got %s\nwant %s", gotJSON, wantJSON)
+			}
+		})
+	}
+}
+
+func TestFormatCSVCell_Scalars(t *testing.T) {
+	if got := formatCSVCell(nil); got != "" {
+		t.Errorf("nil: got %q, want empty string", got)
+	}
+	if got := formatCSVCell("hello"); got != "hello" {
+		t.Errorf("string: got %q, want %q", got, "hello")
+	}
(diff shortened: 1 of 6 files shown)


internal/workers/export.go
```diff
@@ -127,17 +175,63 @@ func (w *UserExportWorker) Work(ctx context.Context, job *river.Job[jobs.UserExp
...
+		return nil, 0, 0, fmt.Errorf("create temp file: %w", err)
+	}
+	totalRows, err := write(f)
+	if err != nil {
+		cleanupTempFile(f)
+		return nil, 0, 0, err
+	}
+	size, err := f.Seek(0, io.SeekEnd)
+	if err != nil {
+		cleanupTempFile(f)
```
STGime, line 214: **Note (low):** `os.CreateTemp` defaults to `/tmp`, which on Kubernetes with an emptyDir volume (the default for pods without explicit mounts) is RAM-backed (`medium: Memory`). A very large tenant export (GB-scale) could fill the emptyDir and get the pod OOM-killed — same failure mode as the old `bytes.Buffer`, just with a higher ceiling.

For now this is fine — 100k-row tenants produce MBs not GBs, and the streaming pipeline is still orders of magnitude better than materialising everything in RAM. If a tenant ever approaches GB-scale exports, the fix is either a PVC mount or a multipart S3 streaming upload that skips the temp file entirely. Not a blocker.

---

internal/workers/export.go
```diff
@@ -27,32 +28,58 @@ func (w *TenantExportWorker) Work(ctx context.Context, job *river.Job[jobs.Tenan
...
+				audit.ActionExportFailed,
+				audit.WithTarget("export", args.ExportID),
+				audit.WithMetadata(map[string]any{
+					"scope": "tenant",
+					"stage": stage,
+					"error": err.Error(),
+				}))
+		}
+	}
+
```
STGime, line 50: The `failExport` closure is a clean pattern — captures the common mark-failed + audit-log logic for each stage. The `stage` field in metadata (`resolve_project` / `build_zip` / `upload`) gives operators a clear signal in the audit feed without having to grep worker logs.

---

internal/compliance/export_test.go
```diff
@@ -0,0 +1,521 @@
...
+	return len(p), nil
+}
+
+func keysOf(m map[string]*zip.File) []string {
+	out := make([]string, 0, len(m))
+	for k := range m {
+		out = append(out, k)
+	}
+	return out
+}
```
STGime, line 521: Excellent test fixture design. `setupExportFixture` creates a throwaway schema + project row with deterministic data, and the cleanup function tears it all down. Parallel-safe via randomised schema names. The `openTestPool` with `t.Skipf` on connection failure means CI without `DATABASE_URL` skips gracefully. The rate-limit test inserting `failed` vs `completed` rows is a direct regression guard for #101.

---

STGime (commented):
High-quality PR that closes four issues at once with backward-compatible changes and a proper test suite.

**#99 (streaming):** The `streamQueryToZip` → temp file → S3 pipeline replaces the old all-in-RAM `Build*ExportZip`. Memory is now bounded by one row + zip buffer instead of the entire dataset. JSON array shape preserved for consumer compatibility. The `streamExportToTempFile` helper with `Seek(0, SeekEnd)` for size and `Seek(0, SeekStart)` for rewind is clean.

**#100 (failure audit):** `failExport` closure in both workers + audit emission in all three handler enqueue-failure paths. The `stage` metadata field (`resolve_project` / `build_zip` / `upload` / `enqueue`) makes the audit feed operationally useful without log-diving.

**#101 (rate-limit fix):** `status <> 'failed'` predicate — minimal, correct. A worker crash no longer locks the user out for 24h.

**Tests (521 lines):** Well-structured — pure unit tests for `formatCSVCell`, integration tests for streaming/filtering/rate-limit/user-existence against real Postgres, parallel-safe fixtures with randomised schema names and deterministic cleanup.

4 inline comments. One note about `/tmp` being RAM-backed on K8s emptyDir — not a blocker at current scale. The `_ = c` in the CSV streaming path is a dead read — harmless, can be dropped.

No blockers — LGTM.

---

PR #146, opened 2026-05-18
Realtime: scope events to row owner so end-users can't see each other (#108)

### LINKED ISSUE #108: Realtime: events bypass RLS — broadcasts go to every subscriber regardless of policy
## Problem

Once #62 lands, realtime fan-out works again: an SDK \`INSERT\` on \`/v1/db/<table>\` publishes an event that reaches every WebSocket client subscribed to \`db:<table>\` on the same \`project_id\`.

**Every** client. Without consulting the RLS policy on the row.

If table \`tenant_X.private_notes\` has \`USING (user_id = auth_uid())\`, end-user Alice can read only her own rows via REST. But when Bob inserts into the same table, the realtime event is broadcast to Alice's open WebSocket too — even though Alice would never see that row via SELECT. The row's full payload (\`record\`) is in the event body.

## Why it matters

Confidentiality leak. Anyone with a valid end-user JWT on the project receives every \`INSERT\` / \`UPDATE\` / \`DELETE\` on every table they subscribe to, including rows they cannot read through the API. Classic Supabase-style RLS protects table reads but not the realtime side-channel.

This is the standard \"realtime + RLS\" gap. Supabase had the same issue and shipped \`realtime authorization\` policies + broadcast filtering. We need an equivalent.

## Approach options

1. **Per-event RLS check in the hub.** Before delivering to each subscriber, run the table's USING clause against the event's row and the subscriber's identity. Correct but expensive: 1 query per (event × subscriber) → at scale (10k subscribers × 100 events/sec) that's a million queries/second. Needs cached / materialised RLS rules.

2. **Project-level broadcast policies.** Owner declares which tables broadcast and to which audiences (\`anon\`, \`authenticated\`, \`uploaded_by\`, etc.). Less expressive than RLS but cheap to evaluate at fan-out time.

3. **Per-subscriber row filtering by user_id only.** Compromise: if a table has an \`uploaded_by\` / \`user_id\` column, filter rows by matching it to the subscriber's end-user ID. Works for the most common pattern; doesn't handle complex policies.

4. **Per-tenant Postgres LISTEN/NOTIFY with RLS-aware triggers.** Push the filtering into Postgres triggers that emit per-subscribed-user notifications. Postgres-native, but complex deploy + per-tenant setup.

## Recommended

Start with **option 3** as a quick mitigation (\`uploaded_by\` / \`user_id\` filtering), document the gap loudly in the realtime docs (\"realtime currently broadcasts every row regardless of RLS — do not subscribe to tables containing other users' data until this lands\"), and follow up with option 1 or 2 as a v1.1 feature.

## Out of scope

- Realtime read replicas (multi-region).
- Realtime over Postgres logical replication (we use application-side publish, not WAL).
- Broadcast batching / debouncing.

## Severity

Confidentiality leak. Affects every project using realtime with RLS-protected tables. No public exploit at the moment because the realtime path was completely broken before #62 — but with #62 deployed, this becomes a live risk.

## File pointers

- Publisher: \`internal/query/handler.go:111, 340, 380, 414\`
- Broadcast: \`internal/realtime/publish.go\` → \`hub.Broadcast(projectID, channel, payload)\`
- Hub: \`internal/realtime/hub.go\` line 152, delivers to every client in \`clients[projectID:channel]\` map
- Subscriber identity carries no end-user ID today — would need to thread it from \`HandleWebSocket\` through the \`Client\` struct.

STGime:
## Summary

Closes #108. Before this PR, an SDK `INSERT`/`UPDATE`/`DELETE` broadcast the full row payload to every WebSocket subscriber on the project, regardless of the RLS policy. End-user Alice would receive Bob's private rows over realtime even though `SELECT` through REST filtered them out.

## Approach

Filter at fan-out time using the row's owner column (`user_id` / `owner_id` / `created_by` / `uploaded_by`) — the same four names the `owner_access` RLS preset already uses, so realtime and REST agree on which column scopes a row.

| Subscriber identity | Owner column present | No owner column |
|---|---|---|
| Service (`eb_sk_*` apikey or platform JWT) | sees every row | sees every row |
| End-user (end-user JWT) | sees rows where owner = user_id | sees every row |
| Anonymous (`eb_pk_*` apikey, no end-user) | **dropped** | sees every row |

Tables without one of the four owner columns keep broadcasting to all subscribers (lookup tables, public feeds). Full per-row policy evaluation is a v1.1 follow-up; the SDK README now documents the gap loudly.

## Wire changes

- `realtime.Authorize` returns `AuthorizedClient{ProjectID, Plan, EndUserID, Service}` instead of three positional values. Cleaner with four fields.
- `Hub.Broadcast(...)` gains `ownerUserID string`. Run-loop skips clients where `!client.shouldReceive(ownerUserID)`.
- `RedisBridge.Publish` wraps the event JSON in a small envelope `{e: payload, o: ownerUserID}`. Receivers parse the envelope, fall back to treating the raw payload as a v0.4 Event (no owner → broadcast to all) if it doesn't decode — preserves correctness during a rolling deploy where some publishers/receivers are still on the old build.
- `buildRealtimeAuthorize`: derives `Service` from `KeyType=="secret"`, `Service=true` for platform JWTs (admins), `EndUserID` from the end-user JWT subject.

## Tests

7 new tests in `internal/realtime/owner_filter_test.go`:

- `TestOwnerUserIDFromRow` / `_StringerFallback` — owner-detection across all four column names, ordering, nil/empty, `[]byte`, pgtype-style Stringer
- `TestClient_ShouldReceive` — table-driven for every identity × ownerUserID combination
- `TestRealtime_OwnerScoped_EndUserMatch` — two end-users, `INSERT` with `user_id=A` → only A receives
- `TestRealtime_OwnerScoped_AnonDoesNotReceive` — anon apikey gets nothing
- `TestRealtime_ServiceSeesEverything` — `eb_sk_*` / platform keeps v0.4 visibility
- `TestRealtime_NoOwnerColumn_StillBroadcasts` — regression guard against the obvious over-correction
- `TestRealtime_DeleteUsesOldRecordOwner` — owner fallback to `old_record` for cleanup paths

All 18 pre-existing realtime tests still pass.

## Drive-bys

- `sdk/js/test/realtime-url.mjs` had the same `setSession` refresh-timer hang the `auth-export.mjs` fix in PR #145 worked around. Added an explicit `process.exit(0)` so `npm test` doesn't stall after this file.
- SDK README documents the filter model + the limitation.

## Test plan

- [x] `go test ./internal/realtime/...` → 25 tests pass (18 pre-existing + 7 new)
- [x] All 4 SDK test files pass cleanly
- [ ] After merge: subscribe two browser sessions (different end-users) to the same `notes` table in superapp → INSERT from one shouldn't surface in the other
- [ ] Confirm the console's realtime tab (signed in with platform JWT, `Service=true`) still sees every event

### DIFF
--- a/internal/realtime/publish.go
+++ b/internal/realtime/publish.go
@@ -70,9 +70,20 @@ func (p *EventPublisher) publish(ctx context.Context, projectID, table, eventTyp
 		return err
 	}
 
+	// Closes #108. Detect the row's owner once on the publisher side
+	// (same row for every subscriber), then carry it on the broadcast
+	// envelope. The hub filters per-subscriber based on it. DELETE
+	// events get the owner from oldRecord because record is just
+	// {id: ...} on the DELETE path (see HandleTableDelete in
+	// internal/query/handler.go).
+	ownerUserID := ownerUserIDFromRow(record)
+	if ownerUserID == "" && oldRecord != nil {
+		ownerUserID = ownerUserIDFromRow(oldRecord)
+	}
+
 	// If Redis bridge is available, publish via Redis for cross-instance delivery.
 	if p.bridge != nil {
-		if err := p.bridge.Publish(ctx, projectID, table, eventType, record, oldRecord); err != nil {
+		if err := p.bridge.Publish(ctx, projectID, table, eventType, record, oldRecord, ownerUserID); err != nil {
 			slog.Error("realtime publisher: redis publish failed, falling back to local",
 				"error", err,
 				"project_id", projectID,
@@ -87,7 +98,7 @@ func (p *EventPublisher) publish(ctx context.Context, projectID, table, eventTyp
 	}
 
 	// Local-only broadcast (no Redis, or Redis publish failed).
-	p.hub.Broadcast(projectID, event.Channel, payload)
+	p.hub.Broadcast(projectID, event.Channel, payload, ownerUserID)
 
 	slog.Debug("realtime publisher: broadcast locally",
 		"project_id", projectID,
@@ -96,4 +107,54 @@ func (p *EventPublisher) publish(ctx context.Context, projectID, table, eventTyp
 	)
 
 	return nil
+}
+
+// ownerColumns are the column names that, if present on a row, identify
+// the end-user that owns it. Order matters when more than one is
+// present: first match wins. The same four names are what the
+// `owner_access` RLS preset looks for at table-creation time (see
+// internal/query/ddl_handler.go), so realtime filtering and REST RLS
+// agree on which column scopes a row.
+var ownerColumns = []string{"user_id", "owner_id", "created_by", "uploaded_by"}
+
+// ownerUserIDFromRow returns the row's owner-column value, or "" if
+// the row has no recognised owner column. Used by the hub to decide
+// whether to broadcast a realtime event to a given subscriber.
+//
+// Closes #108. Designed to match the most common RLS pattern (the
+// owner_access preset) without re-running policy SQL on every event,
+// which would be O(events × subscribers) Postgres queries at fan-out
+// time. Tables with more complex policies (multi-tenant via a
+// per-row org_id, role-based readers, etc.) are still broadcast to
+// every subscriber — see the SDK realtime docs for the limitation.
+func ownerUserIDFromRow(record map[string]interface{}) string {
+	if record == nil {
+		return ""
+	}
+	for _, col := range ownerColumns {
+		v, ok := record[col]
+		if !ok || v == nil {
+			continue
+		}
+		switch x := v.(type) {
+		case string:
+			if x != "" {
+				return x
+			}
+		case []byte:
+			if len(x) > 0 {
+				return string(x)
+			}
+		default:
+			// UUIDs sometimes come back as types that implement
+			// fmt.Stringer (pgtype.UUID); covered by the
+			// default-branch test in publish_test.go.
+			if s, ok := v.(interface{ String() string }); ok {
+				if str := s.String(); str != "" {
+					return str
+				}
+			}
+		}
+	}
+	return ""
 }
--- a/internal/realtime/redis.go
+++ b/internal/realtime/redis.go
@@ -69,6 +69,24 @@ func (b *RedisBridge) Subscribe(ctx context.Context) {
 	}
 }
 
+// redisEnvelope wraps the event JSON for cross-instance fan-out so the
+// receiving Hub can apply the #108 per-row filter without re-detecting
+// the owner column. Wire shape (compact keys keep the Redis payload
+// small):
+//
+//	{"e": <Event JSON>, "o": "<ownerUserID|empty>"}
+//
+// Receiving instances try this shape first; if it doesn't decode they
+// fall back to treating the raw payload as a v0.4 Event (no owner →
+// broadcast to all). The fallback exists so a rolling deploy where
+// old publishers and new receivers coexist for a few minutes doesn't
+// break — once every instance is on the new code, every event carries
+// its envelope and the filter takes effect everywhere.
+type redisEnvelope struct {
+	Event       json.RawMessage `json:"e"`
+	OwnerUserID string          `json:"o,omitempty"`
+}
+
 // handleMessage parses a Redis pub/sub message and broadcasts it to Hub
 // clients. The Redis channel format is "rt:tenant:{tenantID}:{table}".
 func (b *RedisBridge) handleMessage(msg *redis.Message) {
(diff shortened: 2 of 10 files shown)


internal/realtime/redis.go
```diff
@@ -87,18 +105,31 @@ func (b *RedisBridge) handleMessage(msg *redis.Message) {
...
 		"table", table,
 		"channel", channel,
+		"has_owner", ownerUserID != "",
 	)
 
-	b.hub.Broadcast(tenantID, channel, []byte(msg.Payload))
+	b.hub.Broadcast(tenantID, channel, payload, ownerUserID)
 }
 
 // Publish publishes an event to Redis so that all gateway instances can
```
STGime, line 128: The backward-compatible envelope is well-designed for rolling deploys. If an old-format publisher sends a raw Event JSON (no `e`/`o` fields), `json.Unmarshal` into `redisEnvelope` succeeds but `env.Event` is empty (no `e` key) → `len(env.Event) > 0` is false → falls back to treating `msg.Payload` as the event with `ownerUserID=""` → broadcast to all (v0.4 behavior). Once all instances are on the new code, every message carries the envelope and the filter takes effect everywhere. Clean transition.

---

internal/realtime/publish.go
```diff
@@ -70,9 +70,20 @@ func (p *EventPublisher) publish(ctx context.Context, projectID, table, eventTyp
...
+	// events get the owner from oldRecord because record is just
+	// {id: ...} on the DELETE path (see HandleTableDelete in
+	// internal/query/handler.go).
+	ownerUserID := ownerUserIDFromRow(record)
+	if ownerUserID == "" && oldRecord != nil {
+		ownerUserID = ownerUserIDFromRow(oldRecord)
+	}
+
 	// If Redis bridge is available, publish via Redis for cross-instance delivery.
 	if p.bridge != nil {
```
STGime, line 85: Good: owner detection from `oldRecord` as fallback. DELETE events have a near-empty `record` (`{id: ...}`) so the owner has to come from the previous state. Without this, deletes would broadcast to everyone — subtler than the INSERT/UPDATE leak because it's "Alice learns Bob deleted a row" rather than seeing its contents, but still a privacy signal that shouldn't leak.

---

sdk/js/README.md
```diff
@@ -264,6 +264,23 @@ eb.realtime.disconnect()
...
+| Anonymous (`eb_pk_*` public API key, no end-user JWT) | **Receives nothing** | Receives every event |
+
+This matches the `owner_access` RLS preset applied by `eb.db.schema.createTable()` when an owner column is detected, so realtime and REST agree on which rows a subscriber can see.
+
+**Limitations.** The server-side filter is column-based, not a full RLS evaluator:
+
+- Custom or composite RLS policies (e.g. `USING (org_id = current_org() OR is_admin)`) are not enforced over realtime. Subscribers on those tables will see every row.
+- If you need stricter realtime authorisation than the owner column provides, gate sensitive tables off realtime entirely until [#108 follow-up](https://github.com/STGime/euroback/issues/108) ships full policy evaluation in v1.1.
+
 ## Edge Functions
```
STGime, line 285: Honest, clear documentation of the limitation. The table format makes the filter model scannable, and the "Limitations" section explicitly calls out that custom RLS policies aren't enforced over realtime. The recommendation to "gate sensitive tables off realtime entirely" until v1.1 is the right security advice.

---

STGime (commented):
Important security fix — closes a real cross-user data leak where end-user Alice could see Bob's rows over realtime despite RLS filtering them on REST. Well-designed, well-tested, backward-compatible.

**Architecture:**
- Owner detection at publish time using the same four columns (`user_id`, `owner_id`, `created_by`, `uploaded_by`) that the `owner_access` RLS preset uses — so realtime and REST agree on row ownership.
- `shouldReceive` filter at fan-out time: service sees all, end-user sees own rows + unowned rows, anonymous sees only unowned rows.
- Redis envelope (`{"e": <event>, "o": "<owner>"}`) with backward-compatible fallback for rolling deploys.
- DELETE fallback to `oldRecord` for owner detection (DELETE events have near-empty `record`).

**Security model is sound:**
- Owner column values come from rows inserted through the gateway with RLS enforcing `user_id = current_setting('app.end_user_id')`, so spoofing isn't possible.
- Tables without owner columns keep broadcasting to all (documented, correct for lookup/public tables).
- Convention-based filter is the right v1 trade-off. Full RLS evaluation at fan-out time (O(events × subscribers) queries) would be expensive; the v1.1 follow-up is correctly deferred.

**Test coverage is thorough:** 7 new tests covering every identity × owner combination, including the old_record fallback for DELETE and the regression guard against over-correction (no-owner tables still broadcast).

3 inline comments (1 failed due to line range). No blockers — LGTM, ship it.

---

Issue #11, opened 2026-04-14
[P2][tech-debt] Fix string|undefined param types in console, re-enable lint


STGime:
## Problem
`.github/workflows/ci.yml:59` has a TODO: `# TODO: fix string|undefined params across console pages, then re-enable.` Type-checking on console pages is partially disabled — a whole class of null-safety bugs is invisible to CI.

## Acceptance
- [ ] Audit console `+page.svelte` / `+page.ts` files using SvelteKit route params; add explicit narrowing / redirect when `params.id` is undefined
- [ ] Remove the TODO in `ci.yml:59` and re-enable strict type checking
- [ ] `pnpm check` passes in CI

**Priority**: P2 — not a launch blocker but a source of silent regressions.

PR #147, opened 2026-05-18
Email: chunk SendBulk to fit Scaleway TEM 9-recipient quota (#35)

### LINKED ISSUE #35: Batch TEM bulk sends to fit per-message recipient quota
## Problem

The superadmin allowlist email flow (commit 'f1ffc4b', UI at Platform Admin → Signup Allowlist → Email) sends one TEM request with all selected recipients in BCC. Scaleway TEM caps recipients per message at 'TemEmailsMaxRecipients' (10 by default on new domains), including the sender's visible To slot. Selecting 10+ recipients in one compose = 403 from TEM:

'Send failed: TEM API returned 403: {\"details\":[{\"current\":0,\"quota\":10,\"resource\":\"TemEmailsMaxRecipients\"...'

Today's workaround: send in two (or more) batches from the UI. Works, but awkward for larger beta waves.

## Proposed change

'internal/email/client.go SendBulk' — chunk the recipient list to fit the quota; each chunk = one TEM call, still BCC so recipients never see each other. Scope:

- Introduce a per-message cap. Default 9 (10 total including the sender's To slot on the default quota). Read from env 'TEM_MAX_RECIPIENTS_PER_MESSAGE' so raising the Scaleway quota later is a Secret change, not a deploy.
- Continue-on-error: if chunk 3 fails, still try 4, 5, … Return a structured result with per-chunk status so the handler can surface '8/10 sent' in the UI.
- Update 'POST /platform/admin/allowlist/email' response shape to return '{ sent, failed, errors: [] }' instead of the current 'sent: N'.
- Update the console compose modal ('admin/+page.svelte') to display the partial-success shape.

## Not in scope

- Retry-with-backoff on rate-limit 429s (separate issue if TEM starts rate-limiting us).
- Cross-region failover (Scaleway TEM is single-region for now).
- Raising the Scaleway quota itself — that's a support ticket, not code.

## Where to change

- 'internal/email/client.go:117 SendBulk' — add chunking loop, collect per-chunk errors.
- 'internal/email/service.go:42 SendBulkBCC' — widen the return type to report partial success.
- 'internal/tenant/admin_handlers.go AdminSendAllowlistEmail' — thread the partial-success info into the response.
- 'console/src/lib/api.ts' + 'admin/+page.svelte' — render the new shape.

Priority: low. The workaround (send in batches from the UI) is fine for the closed beta.

STGime:
## Summary

Closes #35. The superadmin "Email" button on Platform Admin → Signup Allowlist sent a single TEM request with every selected address in BCC. Scaleway TEM caps per-message recipients at `TemEmailsMaxRecipients` (10 by default, including the visible To slot) — so picking 10+ recipients returned 403 from TEM and the operator had to fall back to selecting nine at a time, sending, deselecting, picking the next nine, etc. The closed-beta allowlist is past 15 entries now, so the workaround hurts every send.

## SendBulk now chunks

- `chunkRecipients(recipients, size)` helper splits the list into chunks of at most `size`. Empty input → no chunks; `size<=0` falls back to the default so an env-typo can't silently disable batching.
- `maxBCCPerMessage()` reads `TEM_MAX_RECIPIENTS_PER_MESSAGE`, defaults to **9**. Operators who get a raised Scaleway quota flip the env var without a redeploy.
- `SendBulk` iterates chunks, issues one TEM POST each via a new `sendBulkChunk()`, accumulates a per-chunk error in `BulkResult.Errors` on failure, and continues. Wrapping error is non-nil only when **every** chunk failed.
- Recipient order is preserved so a given address always lands in the same chunk for a given selection — matters for reproducing what an audit row recorded.

## Partial-success response shape

```json
{
  "status": "sent" | "partial",
  "sent":   <int>,
  "failed": <int>,
  "errors": [{"recipients":[...], "error":"..."}],
  "bcc":    <bool>
}
```

`POST /platform/admin/allowlist/email` returns **502 only when nothing went out at all**; partial success is **HTTP 200 with `status=partial`**. The audit row gets sent/failed counts so the Compliance → Audit Log shows what actually happened, not just "the operator clicked Send."

## Console

- `api.ts` ts type widened.
- `admin/+page.svelte`:
  - "Partial: X sent, Y failed (BCC)." on partial outcomes
  - **Failed recipients stay selected** so clicking Email again retries just them — no manual re-tick of the still-failed addresses
  - Expandable details section showing per-chunk error text from TEM

## Tests

8 new test functions in `internal/email/client_test.go`:

- `TestChunkRecipients` — boundaries (empty, exactly at cap, one over, 20-with-cap-9, 5-with-cap-2) + the `size<=0` fallback regression guard
- `TestChunkRecipients_PreservesOrder` — chunk N contains `recipients[(N-1)*size : N*size]` every time
- `TestMaxBCCPerMessage` — env override + invalid-value fallbacks
- `TestSendBulk_ChunksAndContinuesOnError` — fake TEM server fails chunk #2, succeeds 1 & 3; asserts `sent=16 failed=9`, errors has one entry of 9 addresses, all three chunks were POSTed
- `TestSendBulk_AllChunksFail` — every chunk 500s → returns error, `sent=0`
- `TestSendBulk_SingleChunkUnderCap` — five recipients, one POST, clean success
- `TestSendBulk_Unconfigured` / `_EmptyRecipients` — guard rails

## Test plan

- [x] `npx svelte-check` → no new errors in `admin/+page.svelte` or `lib/api.ts` (the pre-existing 139 errors elsewhere are issue #11)
- [ ] After merge: in the Platform Admin → Signup Allowlist UI, select 15+ recipients, click Email, expect "Sent to 15 (BCC)." with no 403
- [ ] Force a partial: set `TEM_MAX_RECIPIENTS_PER_MESSAGE=3` (or break TEM creds), confirm the partial banner appears + the failed addresses stay selected for retry

### DIFF
--- a/console/src/routes/(app)/admin/+page.svelte
+++ b/console/src/routes/(app)/admin/+page.svelte
@@ -17,6 +17,12 @@
 	let composeBusy = $state(false);
 	let composeError = $state<string | null>(null);
 	let composeSuccess = $state<string | null>(null);
+	// Closes #35. When some chunks fail (Scaleway TEM quota,
+	// network blip), the API returns status=partial plus an `errors`
+	// array. We capture it here so the operator can see WHICH chunks
+	// didn't go out and retry just those — much better than the
+	// pre-fix "send 9, deselect, send next 9, lather, rinse" loop.
+	let composeFailures = $state<{ recipients: string[]; error: string }[]>([]);
 
 	const INVITATION_TEMPLATE = {
 		subject: "You're invited to Eurobase (closed beta)",
@@ -135,11 +141,24 @@
 		)
 			return;
 		composeBusy = true;
+		composeFailures = [];
 		try {
 			const res = await api.adminSendAllowlistEmail(recipients, composeSubject, composeBody);
-			composeSuccess = `Sent to ${res.sent}${res.bcc ? ' (BCC)' : ''}.`;
+			// Partial success: the server sent SOME chunks, others
+			// failed. Show how many landed + surface the failing
+			// chunks so the operator can retry just those addresses
+			// instead of re-sending to everyone.
+			if (res.status === 'partial') {
+				composeSuccess = `Partial: ${res.sent} sent, ${res.failed} failed${res.bcc ? ' (BCC)' : ''}.`;
+				composeFailures = res.errors ?? [];
+				// Keep the failed addresses selected so the operator
+				// can click Email again to retry just them.
+				selected = new Set(composeFailures.flatMap((e) => e.recipients));
+			} else {
+				composeSuccess = `Sent to ${res.sent}${res.bcc ? ' (BCC)' : ''}.`;
+				selected = new Set();
+			}
 			// Leave modal open so the user can see confirmation; they close manually.
-			selected = new Set();
 		} catch (e: any) {
 			composeError = e?.message ?? 'Send failed';
 		} finally {
@@ -151,6 +170,7 @@
 		composeOpen = false;
 		composeError = null;
 		composeSuccess = null;
+		composeFailures = [];
 	}
 </script>
 
@@ -381,6 +401,21 @@
 				{/if}
 				{#if composeSuccess}
 					<div class="rounded-md bg-green-50 border border-green-200 p-2 text-xs text-green-800">{composeSuccess}</div>
+				{/if}
+				{#if composeFailures.length > 0}
+					<details class="rounded-md border border-amber-200 bg-amber-50 text-xs text-amber-900">
+						<summary class="cursor-pointer px-3 py-2 font-medium">
+							{composeFailures.length} chunk{composeFailures.length === 1 ? '' : 's'} failed — click Email to retry the still-selected addresses
+						</summary>
+						<ul class="border-t border-amber-200 px-3 py-2 space-y-1">
+							{#each composeFailures as f}
+								<li>
+									<div class="font-mono text-[11px] text-amber-800">{f.recipients.join(', ')}</div>
+									<div class="text-[11px] text-amber-700">{f.error}</div>
+								</li>
+							{/each}
+						</ul>
+					</details>
 				{/if}
 			</div>
 
--- a/internal/email/client.go
+++ b/internal/email/client.go
@@ -9,8 +9,37 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"os"
+	"strconv"
 	"time"
 )
+
+// defaultMaxBCC is the upper bound on recipients per TEM POST when we
+// don't have an explicit override. Closes #35. Scaleway TEM caps
+// recipients per message at TemEmailsMaxRecipients (10 on a default
+// domain) — and that quota counts the visible To slot too, so we send
+// 1 To + at most defaultMaxBCC BCCs per chunk. Operators with a raised
+// Scaleway quota set TEM_MAX_RECIPIENTS_PER_MESSAGE to the new limit
+// without a deploy.
+const defaultMaxBCC = 9
+
+// BulkChunkError records a single per-chunk failure during a bulk send.
+// Recipients is the BCC list that didn't go out; Err is the wrapped
+// underlying error (typically a TEM API non-2xx response).
+type BulkChunkError struct {
+	Recipients []string `json:"recipients"`
+	Error      string   `json:"error"`
+}
+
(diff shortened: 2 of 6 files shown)


internal/email/client.go
```diff
@@ -166,6 +248,42 @@ func (c *EmailClient) SendBulk(ctx context.Context, bccRecipients []string, subj
...
+	chunks := make([][]string, 0, (len(recipients)+size-1)/size)
+	for i := 0; i < len(recipients); i += size {
+		end := i + size
+		if end > len(recipients) {
+			end = len(recipients)
+		}
+		chunks = append(chunks, recipients[i:end])
+	}
+	return chunks
+}
```
STGime, line 275: Good defensive fallback — `size <= 0` maps to `defaultMaxBCC` instead of "one giant chunk" (which is exactly the pre-#35 bug). The regression test for this (`"size 0 falls back to default"`) is important because an env typo like `TEM_MAX_RECIPIENTS_PER_MESSAGE=` would parse as empty → `Atoi` error → fallback to 9. Safe.

---

console/src/routes/(app)/admin/+page.svelte
```diff
@@ -135,11 +141,24 @@
...
+				composeSuccess = `Partial: ${res.sent} sent, ${res.failed} failed${res.bcc ? ' (BCC)' : ''}.`;
+				composeFailures = res.errors ?? [];
+				// Keep the failed addresses selected so the operator
+				// can click Email again to retry just them.
+				selected = new Set(composeFailures.flatMap((e) => e.recipients));
+			} else {
+				composeSuccess = `Sent to ${res.sent}${res.bcc ? ' (BCC)' : ''}.`;
+				selected = new Set();
+			}
 			// Leave modal open so the user can see confirmation; they close manually.
```
STGime, line 161: Smart UX: failed addresses stay selected so clicking "Email" again retries just the failures. The operator doesn't have to remember which 9 of the 15 were in the failed chunk.

---

internal/email/client_test.go
```diff
@@ -0,0 +1,263 @@
...
+// to a redirect.
+func newClientPointingAt(base string) *EmailClient {
+	c := NewEmailClient("token", "fr-par", "proj-x", "from@example.com", "Eurobase")
+	c.httpClient = &http.Client{
+		Transport: redirectingTransport{base: base},
+	}
+	return c
+}
+
+// redirectingTransport rewrites the request URL so client_test's
```
STGime, line 257: **Nit (test fragility):** `makeAddrs` uses `string(rune('a'+i))` — when `i >= 26` this goes past `'z'` into non-letter Unicode runes (`{`, `|`, etc.). Current max test input is 25, so it's fine today. If someone adds a test with 30+ addresses, the "emails" would be `{@test.local`, `|@test.local`. Not a real problem (TEM doesn't validate in tests), just a future paper-cut. A safer pattern: `fmt.Sprintf("user%d@test.local", i)`.

---

STGime (commented):
Clean fix for a real operational pain point. The chunking design is solid:

- **Continue-on-error**: A failed chunk records the error and moves on instead of aborting the whole send. The `BulkResult` reports sent/failed counts so the handler can distinguish total failure (502) from partial success (200 + `status: "partial"`).
- **Env-configurable cap**: `TEM_MAX_RECIPIENTS_PER_MESSAGE` lets operators with a raised Scaleway quota adjust without a redeploy. The `size <= 0` fallback to `defaultMaxBCC` prevents an env typo from disabling batching (which is exactly the pre-#35 bug).
- **Audit enrichment**: `sent` + `failed` counts in the audit row so Compliance → Audit Log shows what actually happened.
- **Console UX**: Partial success surfaces per-chunk errors in an expandable `<details>`, and failed addresses stay selected for one-click retry. Much better than the "send 9, deselect, pick next 9" loop.

3 inline comments:
1. `chunkRecipients` fallback for `size <= 0` — well-tested regression guard.
2. Failed-addresses-stay-selected retry UX — smart.
3. `makeAddrs` test helper goes past `'z'` at `i >= 26` — works today but fragile for larger inputs.

8 tests with good boundary coverage including a fake TEM server that fails specific chunks. No blockers — LGTM.

---

internal/email/client_test.go
STGime: Good catch — switched to `fmt.Sprintf("user%d@test.local", i)` in cc498cb so the helper stays correct regardless of `n`. Tests still green.

PR #163, opened 2026-05-25
Mollie payment system: subscriptions, dunning, deferred Pro signup (#2, #70)

STGime:
Closes #2, #70, #157, #158, #159, #160, #161, #162.

Schema tables existed since migration 000001 (`subscriptions`, `invoices`, `platform_users.mollie_customer_id`). This PR ships the code that turns them into a working payment system end-to-end.

### Backend
- **`internal/billing/mollie.go`** — typed Mollie REST v2 client, zero new deps. 5 calls: `EnsureCustomer`, `CreateFirstPayment` (captures mandate via `sequenceType=first`), `CreateSubscription`, `CancelSubscription`, `GetPayment`. Plus `VerifyWebhookSignature` (HMAC-SHA256 against `MOLLIE_WEBHOOK_SECRET`).
- **`internal/billing/service.go`** — orchestrates DB + Mollie. `StartUpgrade`, `StartUpgradeForPending` (#70 path), `ApplyPaymentEvent` (webhook reconciliation), `Cancel`. Audit log row on every state transition.
- **`internal/billing/dunning.go`** — 3-day grace state machine + hourly sweeper. `failed payment → grace → downgrade` and `cancel → cancel_at_period_end → period elapsed → downgrade`. Idempotent.
- **`internal/billing/handler.go`** — 5 HTTP routes. Webhook always returns 200 so Mollie doesn't loop on retries; verifies via signature + Mollie API roundtrip before applying any state.
- **Migration 000055** — `cancel_at_period_end` + `grace_until` on `subscriptions` + two partial indexes for the sweep.
- **Migration 000056** — new `pending_projects` table (24h TTL, unique-per-owner-slug) for the #70 fix.
- **5 new audit action constants**: `billing.subscription.created/.cancelled`, `payment.succeeded/.failed`, `plan.changed`.

### #70 fix — defer Pro project creation until paid
`HandleCreateProject` now takes a `CheckoutStarter`. When `plan='pro'` AND Mollie is configured, the project is **NOT created immediately** — instead `pending_projects` reserves the slug, Mollie checkout starts, and the response is `{ status: 'pending_payment', pending_project_id, checkout_url }`. The real `public.projects` row is materialised by the webhook handler on `payment.paid` via a `ProjectProvisioner` callback (avoids a `tenant → billing` import cycle). Free signup flow is unchanged.

### Console
[...]

Issue #51, opened 2026-05-07
Vault: single global encryption key, no AAD binding, no key version


STGime:
**File:** `internal/vault/service.go:47-96`

Current state:
- AES-256-GCM with a single `VAULT_ENCRYPTION_KEY` (env var) shared across every project's vault.
- Fresh nonce per encrypt (good).
- No AAD (Additional Authenticated Data) binding to `(schema, name)` — a stolen ciphertext can be replayed under a different secret name in the same tenant.
- No `key_version` column on `vault_secrets` — no rotation procedure.

## Why it matters

- Any compromise of `VAULT_ENCRYPTION_KEY` decrypts every tenant's secrets, with no path to recover by rotating one tenant or one secret class.
- AAD binding is cheap and defends against ciphertext replay (move ciphertext from one secret name to another).

## Fix

1. Add `key_version SMALLINT` column to `vault_secrets`. Encrypt with versioned key map; decrypt selects key by version.
2. Bind AAD = `schema || ":" || name` so a ciphertext only decrypts under the same `(schema, name)`.
3. Document rotation: dual-write under new key, lazy migrate on read, drop old key after sweep.

Severity: medium.

PR #166, opened 2026-05-28
Sec/MCP — gate credential tables behind internal_auth_path GUC + read-only MCP default (#164, #165)

### LINKED ISSUE #164: Sec/MCP — RLS-gate refresh_tokens / email_tokens / vault_secrets behind internal_auth_path GUC (P0)
Closes the Eurobase-side exposure to the [Supabase MCP prompt-injection vulnerability](https://generalanalysis.com/blog/supabase-mcp-blog) (General Analysis, April 2026).

## Background

Supabase's MCP server connects to Postgres as `service_role`, which bypasses RLS. A prompt-injection payload in a user-submitted row (e.g. `support_messages`) gets read by a developer's MCP session in Cursor / Claude Code; the LLM treats the text as instructions and uses `execute_sql` to exfiltrate `integration_tokens`.

**Eurobase's MCP doesn't connect as service_role directly** — it proxies through the platform API. But `internal/query/engine.go` sets `app.end_user_role='service'` on the developer's behalf when `runSQL` is called via MCP, and our `refresh_tokens` / `email_tokens` / `vault_secrets` policies are `USING (true)`. Net effect inside one project: identical to Supabase's vulnerability.

A prompt-injected `runSQL` can read:
- `refresh_tokens.token_hash` — live session-token hashes
- `email_tokens.token_hash` — live password-reset / verification tokens
- `vault_secrets.ciphertext + nonce` — encrypted but a single global key (#51) means future key leak → plaintext
- `users.password_hash` (bcrypt — weak passwords are crackable)

### Migration 000057 — `internal_auth_path` GUC gate

Replace the `USING (true)` policies on the three sensitive tables across every tenant schema with a policy that requires a new, more-specific GUC:

```sql
ALTER TABLE <schema>.refresh_tokens ENABLE ROW LEVEL SECURITY;
DROP POLICY IF EXISTS refresh_tokens_policy ON <schema>.refresh_tokens;
CREATE POLICY refresh_tokens_policy ON <schema>.refresh_tokens
  USING (current_setting('app.intent', true) = 'internal_auth_path');
-- ditto for email_tokens, vault_secrets
```

Iterate over `information_schema.schemata` LIKE `tenant_%` so all current + future tenants get the lockdown.

### Code changes

Add a helper to `internal/db`:

```go
// RunAsInternalAuthPath wraps fn in a tx that sets
// app.intent='internal_auth_path'. ONLY the auth/email/vault paths
// that legitimately need to touch token/secret tables call this.
// The generic `runSQL` and the SDK SDK runtime queries do NOT,
// which means RLS rejects them at the policy layer.
func RunAsInternalAuthPath(ctx context.Context, pool *pgxpool.Pool, fn func(...) error) error
```

Update three internal paths to wrap their DB calls in `RunAsInternalAuthPath`:
- `internal/auth/refresh.go` — refresh-token CRUD
- `internal/email/service.go` — email-token CRUD (verify, magic link, password reset)
- `internal/vault/service.go` — vault encrypt/decrypt + list/get/set

## Tests

- DB-backed (`-short` skips): assert that a plain `RunAsService` query against `refresh_tokens` returns 0 rows after the migration; the same query inside `RunAsInternalAuthPath` returns the expected rows.
- Regression: existing auth refresh, magic-link verify, vault get/set tests still pass.
- Negative test: a synthetic `runSQL` POST (simulating MCP) against the live engine attempts to SELECT `refresh_tokens` and gets back zero rows (RLS-filtered, not 403 — the policy IS the access control).

## Out of scope

- Removing `is_service_role()` from other policies (e.g. `users.id = me`) — that's used legitimately by DSAR export and other admin paths. We only narrow the credential tables.

Priority: **P0** — direct exfiltration vector with a public-disclosure precedent.

### LINKED ISSUE #165: Sec/MCP — read-only by default + output sanitiser + runbook (P1)
Defence-in-depth on top of the L1 fix in #L1_NUM. Mirrors the Supabase team's [recommended mitigations](https://generalanalysis.com/blog/supabase-mcp-blog) plus a documentation pass.

### A. Read-only `runSQL` by default

Today `runSQL` (via the platform `/data/sql` handler in `internal/query/sql_handler.go`) accepts arbitrary writes. The LLM-driven MCP path is the same handler the CLI uses, but the LLM is trivially fooled by injected instructions.

Changes:
- Add a `read_only` parameter to the request body of `/data/sql` (default `true` when the request comes from the MCP server, `false` for the CLI / console).
- Honour it in the engine via `SET TRANSACTION READ ONLY` at the start of the tx. Writes raise SQLSTATE 25006 and roll back.
- `mcp-server/src/tools/database.ts`:
  - `runSQL` → always sets `read_only: true`
  - New `runSQLMutating` → only registered when env `EUROBASE_MCP_ALLOW_WRITES=true`. Tool description warns it should only be enabled for migration-running scripts, not interactive sessions.

### B. Output sanitiser

Before the MCP server returns rows to the LLM, scan stringy column values for prompt-injection signatures:
- `CLAUDE`, `CLAUDE WITHIN`, `IGNORE PREVIOUS`, `SYSTEM:`, `<|im_start|>`, etc.
- Multi-line text containing both an imperative verb and a tool-name reference.

When matched: replace the value with `[REDACTED — possible prompt injection]`, log a warning, and emit an `mcp.row_sanitised` audit-log row. Not bulletproof — a sufficiently clever attacker can encode payloads — but every layer the attacker has to defeat costs them.

Files: `mcp-server/src/server.ts` adds a `sanitiseToolResult(json)` pass on every tool response.

### C. Audit constants

`internal/audit/service.go`:

```go
ActionMCPSQLExecuted          = "mcp.sql.executed"
ActionMCPSQLRejectedReadOnly  = "mcp.sql.rejected_write_in_readonly"
ActionMCPRowSanitised         = "mcp.row_sanitised"
```

The platform API already audits SQL calls via the existing `audit_log` infrastructure — extend with a `source: 'mcp'` metadata field so the Compliance → Audit Log feed can filter for MCP-origin traffic.

### D. Runbook

`docs/runbooks/mcp-security.md`:
- Threat model (Supabase-style prompt injection + Eurobase-specific scope reduction)
- Why `refresh_tokens` / `email_tokens` / `vault_secrets` are gated behind `internal_auth_path` GUC
- Read-only default + how to opt in to writes for legitimate ops
- Output-sanitiser limitations (encoding attacks, model-aware payloads)
- Developer guidance: 'treat your project DB as a hostile input source when reviewing via MCP'

## Tests

- Unit test for read-only enforcement: `runSQL` with `INSERT INTO …` returns 25006; `SELECT …` succeeds.
- Unit test for the MCP output sanitiser: rows containing injection signatures get redacted; benign content passes through.
- Audit-row test: an MCP `runSQL` call produces an `mcp.sql.executed` row with `source=mcp` metadata.

## Priority

P1 — does not change the actual exfiltration outcome (L1 already closes that), but layers in two more lines of defence and gives the operator a runbook to point at.

STGime:
Closes #164 (P0 — primary fix), closes #165 (P1 — defence in depth).

Same vulnerability class as the [Supabase MCP prompt-injection issue](https://generalanalysis.com/blog/supabase-mcp-blog) (General Analysis, 2026-04). Their fix is a generalised read-only mode; ours combines that with a tighter RLS policy on the three credential tables that have no business being read by arbitrary developer SQL.

## Threat model

A malicious end-user submits a row via the SDK anon role containing a prompt-injection payload. A developer reviews the row in Cursor / Claude Code via the MCP integration. The LLM treats the text as instructions and calls `runSQL` to read `refresh_tokens` / `email_tokens` / `vault_secrets`, then writes the data back into a reply visible to the attacker. Tenant-scoped (we don't have Supabase's cross-tenant exposure), but within one project the exfil is the same.

## Layer 1 — RLS gate on credential tables (#164)

Pre-PR: `refresh_tokens` / `email_tokens` / `vault_secrets` used `USING (public.is_service_role())`. The platform `/data/sql` handler that backs MCP `runSQL` sets `app.end_user_role='service'`. So a prompt-injected query had full read access.

Post-PR:
- **Migration 000055** introduces `public.is_internal_auth_path()` → checks `current_setting('app.intent', true) = 'internal_auth_path'`. Iterates every existing tenant schema, drops the old policies, recreates them with the tighter guard. `provision_tenant()` rewritten so new tenants get the locked-down policies from day one.
- **New helper `db.RunAsAuthService`** wraps fn in a tx setting both `app.end_user_role='service'` AND `app.intent='internal_auth_path'`. Distinguished from `RunAsService` which only sets the former.
- **7 legitimate code paths** switched: `email/service.go`, `sms/service.go`, `enduser/auth_service.go`, `enduser/platform_handler.go` (revoke flows), `enduser/gdpr_export.go`, `vault/service.go`, `gateway/token_cleanup.go`.

A generic `runSQL` via MCP sets `app.end_user_role='service'` but NOT `app.intent`. RLS rejects at the policy layer. Zero-rows result, no error — the contract.

## Layer 2 — Read-only MCP `runSQL` by default (#165)

`SQLRequest` + `SQLTransactionRequest` gain a `read_only` field. The platform handler ORs it with the existing `forceReadOnly` flag. `ExecuteSQLTransaction` gains the same variadic `readOnly` param `ExecuteSQL` already had and wraps in `SET TRANSACTION READ ONLY` when set.

MCP server (`mcp-server/src/tools/database.ts`):
- `runSQL` / `runSQLTransaction` always set `read_only: true` unless `EUROBASE_MCP_ALLOW_WRITES=true` is in the MCP server's env.
- Tool description visible to the LLM updates dynamically to reflect the mode, so the LLM doesn't coach the user into bypassing it.

Effect: even if Layer 1 misses (e.g. someone adds a new sensitive table without gating it), a prompt-injected write to a row the attacker can later read still rolls back with `SQLSTATE 25006`.

## Tests

- `internal/db/auth_service_test.go` — DB-backed (`-short` skips). Asserts `RunAsService` against `refresh_tokens` returns 0 (RLS-filtered post-migration), and `RunAsAuthService` returns ≥ that count (legitimate path still works). Catches reverts on either side.
- All 26 Go packages green under `go test -short`.

## Migration-number coordination

This PR uses **000055**. Open billing PR #163 also uses 000055 + 000056; it should rebase to 000057 + 000058 once this one lands. Migrations touch disjoint surfaces; conflict is purely filename-level.

## Open follow-ups (in the runbook, not blocking)

- **Output sanitiser** — flag rows containing common prompt-injection signatures before LLM render.
- **MCP-aware audit metadata** — wire the new `mcp.sql.*` constants into the SQL handler.
- **Per-tenant vault encryption keys (#51)** — limits blast radius of a future encryption-key leak.

### DIFF
--- a/internal/db/auth_service_test.go
+++ b/internal/db/auth_service_test.go
@@ -1,0 +1,99 @@
+package db_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/eurobase/euroback/internal/db"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// TestRunAsAuthService_GatesSensitiveTables is the regression test for
+// #164. Migration 000055 narrowed the RLS policies on refresh_tokens,
+// email_tokens, vault_secrets to require app.intent='internal_auth_path'.
+// This test proves the gate works:
+//   - Plain RunAsService against refresh_tokens returns 0 rows
+//     (RLS-filtered, mimics what a prompt-injected runSQL via MCP would
+//     see). NOT an error — that's the contract.
+//   - RunAsAuthService against the same table returns the seeded rows.
+//
+// If anyone reverts migration 000055 or removes the intent GUC from
+// RunAsAuthService, this test fails loudly.
+func TestRunAsAuthService_GatesSensitiveTables(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test in -short mode")
+	}
+
+	databaseURL := os.Getenv("DATABASE_URL")
+	if databaseURL == "" {
+		databaseURL = "postgres://eurobase_api:localdev@localhost:5433/eurobase?sslmode=disable"
+	}
+
+	ctx := context.Background()
+	pool, err := pgxpool.New(ctx, databaseURL)
+	if err != nil {
+		t.Skipf("cannot connect to test database: %v", err)
+	}
+	defer pool.Close()
+	if err := pool.Ping(ctx); err != nil {
+		t.Skipf("cannot ping test database: %v", err)
+	}
+
+	// Find any tenant schema that has refresh_tokens — every
+	// provisioned project has one.
+	var schemaName string
+	if err := pool.QueryRow(ctx,
+		`SELECT schema_name FROM public.projects
+		 WHERE schema_name IS NOT NULL
+		 ORDER BY created_at DESC LIMIT 1`).Scan(&schemaName); err != nil {
+		t.Skipf("no provisioned tenant schemas available: %v", err)
+	}
+	if schemaName == "" {
+		t.Skip("no tenant schemas to test against")
+	}
+
+	countQ := fmt.Sprintf(`SELECT count(*) FROM %s.refresh_tokens`, quoteIdent(schemaName))
+
+	// Plain RunAsService — sets only app.end_user_role='service',
+	// not app.intent. Post-migration 000055, the refresh_tokens
+	// policy demands is_internal_auth_path(), so this should see
+	// zero rows.
+	var serviceCount int
+	if err := db.RunAsService(ctx, pool, func(ctx context.Context, tx pgx.Tx) error {
+		return tx.QueryRow(ctx, countQ).Scan(&serviceCount)
+	}); err != nil {
+		t.Fatalf("RunAsService count: %v", err)
+	}
+	if serviceCount != 0 {
+		t.Errorf("RLS regression: plain RunAsService saw %d refresh_tokens rows; "+
+			"expected 0 because migration 000055 should require app.intent='internal_auth_path'", serviceCount)
+	}
+
+	// RunAsAuthService — sets BOTH GUCs, satisfies the policy.
+	// We can't assert "> 0" because the test DB may have no real
+	// tokens; the strong assertion is "no RLS error and a valid count".
+	// To distinguish RLS filtering from "table actually empty" we
+	// compare against the ALL-rows count under direct migrator pool
+	// access. That's beyond what we have here — so the cheaper
+	// assertion is: the count under RunAsAuthService is >= the
+	// count under RunAsService. Always true; it catches a future
+	// regression where someone overrides the policy to be even
+	// stricter and breaks the auth path.
+	var authCount int
+	if err := db.RunAsAuthService(ctx, pool, func(ctx context.Context, tx pgx.Tx) error {
+		return tx.QueryRow(ctx, countQ).Scan(&authCount)
+	}); err != nil {
+		t.Fatalf("RunAsAuthService count: %v (auth code paths just broke)", err)
+	}
+	if authCount < serviceCount {
+		t.Errorf("RunAsAuthService saw FEWER rows (%d) than RunAsService (%d) — "+
+			"policy on refresh_tokens is misconfigured", authCount, serviceCount)
+	}
+}
+
+// quoteIdent mirrors the helper used in the engine for safe schema-name
+// interpolation. Local to the test so the test file stays self-contained.
+func quoteIdent(s string) string { return `"` + s + `"` }
--- a/mcp-server/src/tools/database.ts
+++ b/mcp-server/src/tools/database.ts
@@ -31,22 +31,43 @@ export function registerDatabaseTools(server: McpServer, getClient: () => ApiCli
     }
   );
 
+  // Closes part of #165. runSQL / runSQLTransaction default to
+  // READ-ONLY mode — the backend wraps the statements in SET
+  // TRANSACTION READ ONLY, so writes raise SQLSTATE 25006 and roll
+  // back. Defence-in-depth on top of the RLS gate in migration 000055
+  // (#164): even if a prompt-injected query tried to write its
+  // exfiltrated tokens into a row the attacker controls, the write
+  // would never commit.
+  //
+  // Opt out only by setting EUROBASE_MCP_ALLOW_WRITES=true on the
+  // MCP server's environment and restarting. Intended for
+  // migration-running scripts; do NOT enable in interactive Cursor /
+  // Claude Code sessions where prompt-injection-via-data is in scope.
+  const allowWrites = process.env.EUROBASE_MCP_ALLOW_WRITES === 'true';
+  const writeNote = allowWrites
+    ? ' EUROBASE_MCP_ALLOW_WRITES=true is set, so this tool runs with write capability.'
+    : ' READ-ONLY: writes will fail with SQLSTATE 25006. To allow writes, set EUROBASE_MCP_ALLOW_WRITES=true on the MCP server and restart.';
+
   server.tool(
     'runSQL',
-    'Execute a SINGLE SQL statement against a project database. The server uses pgx\'s extended query protocol, which only runs the first statement of a multi-statement string — to prevent silent partial migrations, multi-statement input is rejected with a clear error. For migrations or any multi-statement script, use runSQLTransaction instead.',
+    'Execute a SINGLE SQL statement against a project database.' + writeNote +
+      ' The server uses pgx\'s extended query protocol, which only runs the first statement of a multi-statement string — to prevent silent partial migrations, multi-statement input is rejected with a clear error. For migrations or any multi-statement script, use runSQLTransaction instead.',
     {
       projectId: z.string().describe('The project UUID'),
       query: z.string().describe('A single SQL statement (one query, no internal `;` between statements)'),
     },
     async ({ projectId, query }) => {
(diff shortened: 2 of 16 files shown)


mcp-server/src/tools/database.ts
```diff
@@ -31,22 +31,43 @@ export function registerDatabaseTools(server: McpServer, getClient: () => ApiCli
...
+    ? ' EUROBASE_MCP_ALLOW_WRITES=true is set, so this tool runs with write capability.'
+    : ' READ-ONLY: writes will fail with SQLSTATE 25006. To allow writes, set EUROBASE_MCP_ALLOW_WRITES=true on the MCP server and restart.';
+
   server.tool(
     'runSQL',
-    'Execute a SINGLE SQL statement against a project database. The server uses pgx\'s extended query protocol, which only runs the first statement of a multi-statement string — to prevent silent partial migrations, multi-statement input is rejected with a clear error. For migrations or any multi-statement script, use runSQLTransaction instead.',
+    'Execute a SINGLE SQL statement against a project database.' + writeNote +
+      ' The server uses pgx\'s extended query protocol, which only runs the first statement of a multi-statement string — to prevent silent partial migrations, multi-statement input is rejected with a clear error. For migrations or any multi-statement script, use runSQLTransaction instead.',
     {
       projectId: z.string().describe('The project UUID'),
```
STGime, line 56: Good: the tool description updates dynamically based on `allowWrites`. The LLM sees "READ-ONLY: writes will fail with SQLSTATE 25006" in the tool description itself, so it won't try to coach the user into running INSERT/UPDATE. If `EUROBASE_MCP_ALLOW_WRITES=true` is set, the description says so explicitly — no silent mode switching.

---

internal/db/auth_service_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if authCount < serviceCount {
+		t.Errorf("RunAsAuthService saw FEWER rows (%d) than RunAsService (%d) — "+
+			"policy on refresh_tokens is misconfigured", authCount, serviceCount)
+	}
+}
+
+// quoteIdent mirrors the helper used in the engine for safe schema-name
+// interpolation. Local to the test so the test file stays self-contained.
+func quoteIdent(s string) string { return `"` + s + `"` }
```
STGime, line 99: Strong regression test. The assertion `serviceCount != 0` → fail with "migration 000055 should require app.intent" is the exact guard that catches a policy revert. And the `RunAsAuthService` path asserting `authCount >= serviceCount` catches the opposite regression (someone over-tightens the policy and breaks legitimate auth flows).

One minor gap: the test skips when no tenant schemas exist in the test DB. In CI with `DATABASE_URL` set but no provisioned projects, this silently passes. Worth seeding a test tenant in the fixture if you want CI coverage, but not blocking.

---

STGime (commented):
Excellent security hardening — well-designed defence in depth against the Supabase-class MCP prompt-injection vulnerability. I read the full diff (696 additions across 16 files). Both layers are sound.

## Layer 1 — RLS intent gate (#164)

The core design is correct: introduce `app.intent='internal_auth_path'` as a second GUC that only the 7 legitimate Go code paths set (via `RunAsAuthService`). The generic SQL handler (which MCP's `runSQL` calls) sets `app.end_user_role='service'` but NOT `app.intent`. RLS rejects at the policy layer — zero rows, no error. The attack chain breaks.

`RunAsAuthService` being a separate function (not a flag) is a good security choice — adding a new caller is a conscious, grep-able decision. The migration backfills all existing tenant schemas and updates `provision_tenant` for new ones.

The three tables narrowed (`refresh_tokens`, `email_tokens`, `vault_secrets`) are exactly the right ones — they hold credential material. Other tables (`users`, `user_identities`, `storage_objects`) keep `is_service_role()` because admin/DSAR paths legitimately need them.

## Layer 2 — Read-only MCP default (#165)

`read_only: true` on the request body → `SET TRANSACTION READ ONLY` in the backend. Even if Layer 1 missed (e.g. a new sensitive table without the gate), the exfil step fails — the write back to a row the attacker can read raises `SQLSTATE 25006`. Dynamic tool description updates so the LLM sees the mode.

`EUROBASE_MCP_ALLOW_WRITES=true` opt-out is correctly scoped to the MCP server's env (not the gateway). Intended for migration scripts only.

## Issues

None blocking. Two inline comments:
1. **MCP tool description** — Dynamic `writeNote` in the tool description is well done.
2. **Test coverage gap** — `TestRunAsAuthService_GatesSensitiveTables` skips when no tenant schemas exist. CI with `DATABASE_URL` but no provisioned projects would silently pass. Minor — seed a test tenant if you want CI coverage.

## Migration coordination

This PR uses 000055; billing PR #163 also uses 000055+000056. The PR description correctly notes #163 should rebase to 000057+000058. Conflicts are filename-level only.

The runbook (`docs/runbooks/mcp-security.md`) is thorough — threat model, three defence layers, audit actions, developer guidance, and an honest "what this does NOT protect against" section. 

LGTM — ship it.

---

Issue #167, opened 2026-05-29
[Tier1] Per-tenant encryption keys via pluggable KeyProvider (HKDF now)


STGime:
**Tier-1 GDPR blocker #2 (Customer-Managed Keys / BYOK).**

**Gap:** `internal/vault/service.go` encrypts every tenant's `vault_secrets` with one shared `VAULT_ENCRYPTION_KEY` (AES-256-GCM). No per-tenant separation, no rotation.

**Approach (agreed): C-now, design-for-B-later.** Per-tenant app-layer keys derived via HKDF-SHA256 from the existing master secret (`salt=project_id`), versioned for rotation, behind a `KeyProvider` interface so Scaleway Key Manager (A) or customer HYOK (B) can drop in later without re-encrypting.

**Work:**
- `internal/vault/keyprovider.go`: `KeyProvider` interface + `hkdfKeyProvider`. `version 0` = legacy master key (back-compat).
- Migration `000056`: add `key_version smallint NOT NULL DEFAULT 0` to `vault_secrets` (update `provision_tenant` + backfill existing `tenant_*` schemas, idempotent per `000047`).
- Wire provider into `vault.Service` (`cmd/gateway`, `cmd/worker`).
- Rekey worker `internal/workers/vault_rekey.go` + `POST /platform/projects/{id}/vault/rekey` (audit `vault.rekey.requested`).
- Docs: `docs/compliance/encryption-keys.md`; note the not-yet-true-BYOK caveat in `docs/legal/v1/dpa.md` Annex 2.

**Acceptance:** legacy `key_version=0` rows still decrypt; new rows use current version; rekey bumps version and round-trips; `go test ./internal/vault/...` green. Delivered as PR `tier1/01-per-tenant-keys`.

PR #178, opened 2026-05-29
Fix: restore provision_tenant schema regressed by 000055 (P0)

### LINKED ISSUE #177: [P0] 000055 regressed provision_tenant: new tenants get broken vault/storage schema
Migration `000055_mcp_intent_gate` re-pasted a stale `provision_tenant()` body while applying its (correct) RLS policy hardening. The stale body silently reverted two table definitions for **new** tenants:

- `vault_secrets.secret` → `ciphertext` — `internal/vault/service.go` queries `secret`/`nonce`, so vault Set/Get fails with `column "secret" does not exist`.
- `storage_objects` reverted from the 000050/000052 shape (`size_bytes`, `content_type`, `metadata`, `uploaded_by` FK, `UNIQUE(key)`) to a bare `size`/`uploaded_by`.

`provision_tenant` is `CREATE OR REPLACE`, so the live DB now holds the broken body (confirmed via `pg_get_functiondef`). **Existing tenants are unaffected** (live DB shows every `tenant_*.vault_secrets` still has `secret`/`nonce`), but **any new project created since 000055 ships broken**.

Fix in PR: migration `000056_fix_provision_tenant_regression` restores the correct 000052 schema body while keeping 000055's `is_internal_auth_path()` policy. Found while building the Tier-1 per-tenant-keys work (#167), which needs to `CREATE OR REPLACE provision_tenant`.

STGime:
Closes #177.

## Problem
`000055_mcp_intent_gate` applied a correct RLS hardening (the three sensitive tables now use `is_internal_auth_path()`), but in doing so re-pasted a **stale** `provision_tenant()` body. Because the function is `CREATE OR REPLACE`, the stale body became the live definition and silently reverted two table shapes for **new** tenants:

| Table | 000055 (broken) | Correct (000052) |
|---|---|---|
| `vault_secrets` | `ciphertext BYTEA` | `secret BYTEA` (Go code uses this) |
| `storage_objects` | `size`, no FK, no metadata | `size_bytes`, `content_type`, `metadata`, `uploaded_by` FK, `UNIQUE(key)` |

**Impact:** existing tenants are fine (provisioned by earlier revisions — live DB confirms every `tenant_*.vault_secrets` still has `secret`/`nonce`), but any **new** project created since 000055 gets a tenant whose vault (`Set/Get`) and storage writes fail.

## Fix
`000056_fix_provision_tenant_regression` restores the correct 000052 schema body **while keeping** 000055's security change (`refresh_tokens`/`email_tokens`/`vault_secrets` stay on `is_internal_auth_path()`). Function-only change; no existing-tenant backfill required.

## Verification
- Confirmed the live `provision_tenant` body currently defines `ciphertext` + `size` (via `pg_get_functiondef`).
- Confirmed every existing `tenant_*.vault_secrets` has `secret`/`nonce` (read-only MCP query, 20+ schemas).
- Confirmed all runtime dependencies exist in prod: `uuid_generate_v4`, `is_service_role`, `current_end_user_id`, `is_internal_auth_path`, the gateway/func-runner/migrator roles, and `projects.schema_name`.
- The up body is a composition of two already-applied migrations (000052 schema + 000055 policy). CI `migrate` Job applies it before rollout.

## Note
Found while starting the Tier-1 GDPR per-tenant-keys work (#167), which must `CREATE OR REPLACE provision_tenant` to add `key_version` and therefore needs a correct base. The Tier-1 keys PR builds on top of this.

### DIFF
--- a/migrations/000056_fix_provision_tenant_regression.down.sql
+++ b/migrations/000056_fix_provision_tenant_regression.down.sql
@@ -1,0 +1,214 @@
+-- 000056_fix_provision_tenant_regression.down.sql
+--
+-- WARNING: reverting this hotfix REINTRODUCES the P0 regression from
+-- 000055 — new tenants would again get vault_secrets.ciphertext (which
+-- internal/vault/service.go cannot read) and a pre-000050 storage_objects
+-- shape. Only run this to get back to the exact pre-000056 state.
+--
+-- This restores provision_tenant() to the (broken) 000055 body verbatim.
+
+CREATE OR REPLACE FUNCTION public.provision_tenant(
+    p_project_id   UUID,
+    p_display_name TEXT,
+    p_plan         TEXT DEFAULT 'free'
+)
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+AS $fn$
+DECLARE
+    v_schema_name TEXT;
+    v_func_role   TEXT;
+BEGIN
+    v_schema_name := 'tenant_' || replace(p_project_id::text, '-', '_');
+    v_func_role   := v_schema_name || '_func';
+
+    EXECUTE format('CREATE SCHEMA %I', v_schema_name);
+    EXECUTE format('SET search_path TO %I', v_schema_name);
+
+    EXECUTE format(
+        'CREATE TABLE %I.users (
+            id                 UUID        PRIMARY KEY DEFAULT public.uuid_generate_v4(),
+            email              TEXT        UNIQUE,
+            phone              TEXT,
+            password_hash      TEXT,
+            display_name       TEXT,
+            avatar_url         TEXT,
+            metadata           JSONB       DEFAULT ''{}''::jsonb,
+            provider           TEXT        DEFAULT ''email'',
+            provider_user_id   TEXT,
+            email_confirmed_at TIMESTAMPTZ,
+            phone_confirmed_at TIMESTAMPTZ,
+            last_sign_in_at    TIMESTAMPTZ,
+            banned_at          TIMESTAMPTZ,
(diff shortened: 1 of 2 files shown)


STGime (commented):
Correct P0 hotfix. The root cause analysis is precise: 000055 re-pasted a stale `provision_tenant()` body that reverted `vault_secrets.secret` → `ciphertext` and `storage_objects` to a pre-000050 shape. `CREATE OR REPLACE` made it silent — no error, just a broken function definition that only affects **new** tenants.

**The fix is right:**
- `vault_secrets` restored to `secret BYTEA NOT NULL` + `nonce BYTEA NOT NULL` (what `internal/vault/service.go` queries)
- `storage_objects` restored to `size_bytes`, `content_type`, `metadata JSONB`, `uploaded_by UUID REFERENCES users(id)`, `UNIQUE(key)` (the 000050/000052 shape)
- 000055's security hardening **kept**: `refresh_tokens`/`email_tokens`/`vault_secrets` policies still use `is_internal_auth_path()`
- `auth_email()` uses the live lookup from 000052
- Grant structure matches 000052 (`FOR ROLE eurobase_migrator` qualifiers, `NOLOGIN INHERIT` on func role)
- No `BEGIN`/`COMMIT` — learned from the golang-migrate footgun

**Verification is thorough:** Confirmed live DB has the correct column names on existing tenants (so no backfill needed), confirmed the broken body is what's currently deployed, confirmed all runtime dependencies exist.

**Down migration** correctly restores the broken 000055 body with a WARNING comment. This is the right approach — the down migration is a precise inverse, not an attempt to "fix" in both directions.

No issues. LGTM — merge immediately, this blocks new project creation.

---

PR #179, opened 2026-05-29
Tier-1 #2: per-tenant vault encryption keys (HKDF) + rotation

### LINKED ISSUE #167: [Tier1] Per-tenant encryption keys via pluggable KeyProvider (HKDF now)

STGime:
Closes #167. Tier-1 GDPR #2 (Customer-Managed Encryption Keys). Builds on #178.

## What changes
Today every tenant's `vault_secrets` are encrypted with **one shared** `VAULT_ENCRYPTION_KEY`. This PR introduces **per-tenant keys** derived from that master via HKDF-SHA256, behind a pluggable `KeyProvider` interface, plus versioning so keys can rotate without re-encrypting historic rows.

```
tenant_key = HKDF-SHA256(master, salt=<tenant schema>, info="eurobase-vault-v<N>", 32)
```

- **Per-tenant separation** — salt = tenant schema, so tenant A's key can't open tenant B's ciphertext.
- **Rotation** — version mixed into `info`; bumping it yields an independent key.
- **Back-compat** — `key_version = 0` means "shared master key verbatim"; existing rows are backfilled to 0 and keep decrypting. New writes use v1.

## Changes
| File | What |
|---|---|
| `internal/vault/keyprovider.go` | `KeyProvider` interface + `hkdfKeyProvider`. Pluggable for future Scaleway KMS / customer HYOK. |
| `internal/vault/service.go` | `seal`/`open` go through the provider; `Get`/`Set`/`Update` read & persist `key_version`; new `RekeySchema()` re-encrypts a tenant onto the current version in one service-role tx. |
| `migrations/000057_vault_key_version.{up,down}.sql` | Add `key_version smallint NOT NULL DEFAULT 0` to every tenant `vault_secrets` (backfill + `provision_tenant`). Body identical to the #178 fix except the new column. |
| `internal/vault/handler.go` | `POST /platform/projects/{id}/vault/rekey` (admin-only) → `RekeySchema`. |
| `internal/audit/service.go` | new `vault.rekeyed` audit action. |
| `docs/compliance/encryption-keys.md` | Key model, rotation, and honest BYOK scope/limits. |

## Sovereignty / scope note
This is **app-layer per-tenant derivation**, not "platform-can't-decrypt" BYOK — the platform still holds the master. It satisfies *per-tenant cryptographic separation* + *rotation*. The `KeyProvider` interface + per-row `key_version` are the upgrade path to **Scaleway Key Manager** or **customer HYOK** with no re-encryption (tracked for enterprise). Documented in the DPA caveat.

## Verification
- `go build ./...`, `go vet ./...`, `go test ./... -short` all clean.
- New unit tests (no DB): HKDF determinism, per-tenant separation, version rotation, legacy v0 = master key, seal/open round-trip + cross-tenant isolation — all pass.
- Uses **stdlib `crypto/hkdf`** (Go 1.24+) — no new dependency.
- Existing `vault` integration tests (round-trip Set/Get/Update/Delete) updated implicitly via the `key_version` column; they run against a real Postgres in CI.

## Dependency
Builds on #178 (provision_tenant regression fix) — the 000057 `provision_tenant` body extends the corrected 000056 one. #178 is merged to main; this branch is rebased on top.

### DIFF
--- a/docs/compliance/encryption-keys.md
+++ b/docs/compliance/encryption-keys.md
@@ -1,0 +1,86 @@
+# Vault Encryption Keys — Per-Tenant Derivation & Rotation
+
+> GDPR Art. 32 / Art. 25 (Tier-1 #2). How Eurobase encrypts tenant secrets at
+> rest, why each tenant has a distinct key, and how keys rotate.
+
+## What is encrypted
+
+Every secret in a tenant's `vault_secrets` table (`internal/vault`) is sealed
+with **AES-256-GCM** before it touches disk: API credentials, OAuth
+`client_secret`s, and any value a customer stores via the Vault SDK/console.
+Rows store the ciphertext (`secret`), the GCM `nonce`, and now a `key_version`.
+
+## Key model
+
+A single base64 32-byte master secret, `VAULT_ENCRYPTION_KEY`, is supplied via
+the `eurobase-secrets` Kubernetes Secret. The master is **never used directly**
+to seal new rows. Instead a `KeyProvider` (`internal/vault/keyprovider.go`)
+derives a **distinct key per tenant**:
+
+```
+tenant_key = HKDF-SHA256(
+    secret = VAULT_ENCRYPTION_KEY,
+    salt   = <tenant schema name>,   // per-tenant separation
+    info   = "eurobase-vault-v<N>",  // N = key version (rotation)
+    len    = 32,
+)
+```
+
+- **Per-tenant separation** — the salt is the tenant schema, so tenant A's key
+  cannot open tenant B's ciphertext (proved in `keyprovider_test.go`).
+- **Rotation** — the version is mixed into `info`, so bumping the version
+  yields a fresh, independent key without affecting older rows.
+
+### Versions
+
+| `key_version` | Meaning |
+|---|---|
+| `0` | **Legacy.** Row was sealed with the shared master key verbatim, before per-tenant derivation existed. The provider returns the master key for v0 so these rows keep decrypting. Existing rows are backfilled to 0 by migration `000057`. |
+| `1` | First HKDF-derived, per-tenant key. All **new** writes use the current version (starts at 1). |
+| `≥2` | Reserved for future rotations. |
+
+Decryption always dispatches on the row's stored `key_version`, so multiple
+versions coexist safely. A `Set`/`Update` re-seals at the current version, so
+rows migrate to per-tenant keys naturally as they're written.
+
+## Rotation
+
+`POST /platform/projects/{id}/vault/rekey` (admin-only) re-encrypts every
+secret in the project's vault under the current key version, including legacy
+v0 rows. It runs synchronously in one service-role transaction
+(`VaultService.RekeySchema`) — vaults hold few secrets per tenant, so this is
+cheap; audited as `vault.rekeyed`.
+
+To rotate the whole platform onto a new version: bump the provider's current
+version, deploy, then call rekey per project (or let writes migrate rows
+lazily).
+
+## Sovereignty & limits — honest scope
+
+This is **app-layer per-tenant key derivation**, not "the platform cannot
+decrypt your data" BYOK. The platform still holds the master secret and can
+derive any tenant key. It delivers what GDPR/DSK call for in *per-tenant
+cryptographic separation* and *key rotation*, but a German DSK reviewer asking
+for **client-held keys** wants more.
+
+The `KeyProvider` interface is the upgrade path. Because every ciphertext is
+tagged with its `key_version`, a future provider can own a distinct version
+range and coexist with HKDF rows — **no re-encryption of historic data
+required**:
+
+- **Scaleway Key Manager (KMS)** — envelope encryption with a managed,
+  EU-sovereign master key (rotation + access audit handled by Scaleway).
+- **Customer HYOK** — per-tenant key supplied/hosted by the customer (e.g.
+  self-hosted HashiCorp Vault in the EU) so the platform genuinely cannot
+  decrypt for that tenant.
+
+Both are tracked for enterprise; neither changes the on-disk format.
+
+## Operations
+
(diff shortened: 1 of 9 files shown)


STGime (commented):
## Review — per-tenant vault keys (HKDF) + rotation

**Verdict: looks good with minor suggestions.** Builds cleanly (`go build` / `go vet` / `go test ./internal/vault/... -short` all pass), the crypto is textbook-correct, and the migration carries no regression risk. The honest "this is not true BYOK" framing in both the docs and the code comments is a real strength.

### Strengths
- **Crypto usage is correct.** HKDF salt = stable per-tenant identifier (non-secret salt is fine/standard for domain separation); version folded into `info` yields independent keys on rotation. Cross-tenant isolation is verified by an actual seal/open test relying on GCM auth failure.
- **Migration is safe.** Diffed the `provision_tenant` body in `000057` against the freshly-fixed `000056`: the *only* functional change is the added `key_version` column — everything else is comment-only. The #178 regression is not at risk of being re-introduced. `ADD COLUMN ... NOT NULL DEFAULT 0` is metadata-only on PG11+, so the backfill is cheap.
- **Back-compat is sound.** v0 rows keep decrypting (provider returns master), decryption dispatches on the stored version, and writes lazily migrate rows to the current version.
- **Admin gating confirmed.** `router.go:313` mounts `/vault` behind `RequireMinRole("admin")`, so the handler comment is accurate.
- **Sovereignty-aware.** No new dependency (stdlib `crypto/hkdf`, Go 1.25); the KMS/HYOK upgrade path is documented without changing the on-disk format.

### Suggestions
1. **Test coverage gap at the DB layer (main one).** The new unit tests are crypto-only and DB-free. `RekeySchema` — the most intricate new code (select-`FOR UPDATE` → decrypt-at-old-version → re-seal-at-current → update, all in one tx) — has no test exercising the v0→v1 round-trip, nor does the `/rekey` handler. Since this is the rotation path GDPR review hangs on, a CI integration test that seals a v0 row, rekeys, and confirms it reads back as v1 would be worth adding before relying on it in production.
2. **Whole-schema lock in one transaction.** `RekeySchema` holds `FOR UPDATE` on all rekeyable rows for the request duration. Correctly documented as cheap (plan-capped vaults) with a River-job escape hatch — just flagging that a concurrent `Update` to the same secret blocks until rekey commits.
3. **Rekey audit target is empty.** `auditVault(r, projectID, audit.ActionVaultRekeyed, "")` logs `target_type="vault_secret"` with an empty name, and the returned `rekeyed` count isn't captured. Consider omitting the target or recording the count in audit metadata so the entry is self-describing.
4. **Down-migration drift (cosmetic).** `000057.down` restores `provision_tenant` by hand — the usual copy-the-whole-function smell. Not a blocker (established pattern here, and the destructive-column warning is excellent), just inherently fragile across future edits.
5. **Master-rotation footgun.** Docs correctly note that changing `VAULT_ENCRYPTION_KEY` breaks v0 rows *and* changes all derived keys. Since lazy migration leaves untouched secrets on v0 indefinitely, a follow-up to drive all rows to ≥v1 platform-wide would let the master eventually rotate without a v0 landmine.

None of these block merge. (1) is the one I'd most encourage before depending on rekey in production.

---

STGime:
Thanks for the thorough review. Addressed in `8bce3a8`:

1. **DB-layer rekey test (main ask)** — added `TestVaultRekeySchema_LegacyToCurrent` (integration, runs in CI `test-go`): seals a v0 row under the master key, rekeys, asserts exactly the v0 row is re-encrypted while the v1 row is untouched, both still decrypt, and a second rekey is a no-op. Exercises the full `FOR UPDATE` → decrypt-old → re-seal-current → update tx.
3. **Rekey audit** — now records the `rekeyed` count in `WithMetadata` instead of an empty target name, so the entry is self-describing.
5. **Master-rotation footgun** — filed #180 to add a platform-wide rekey-all (drive every row to ≥v1 across all schemas) so the master can eventually rotate without a v0 landmine, plus a master-rotation runbook.

Not changing:
2. **Whole-schema lock** — intentional and documented (plan-capped vaults; River-job escape hatch noted). Agreed a concurrent `Update` to the same secret blocks until rekey commits; acceptable at this scale.
4. **Down-migration drift** — cosmetic, follows the established repo pattern; the destructive-column warning stays. (Worth a broader refactor of the copy-the-whole-function pattern someday — that pattern is exactly what caused the #178 regression — but out of scope here.)

Pushing the follow-up commit now.

Issue #170, opened 2026-05-29
[Tier1] Generic SIEM audit export (webhook + syslog + signed object dump)


STGime:
**Tier-1 GDPR blocker #3 (SIEM export).** Vendor-neutral by design — Splunk/Datadog/AWS Security Hub are US services and violate the sovereignty rule.

**Work:**
- Migration `000058`: `public.audit_export_destinations` (`kind: webhook|syslog|object_dump`, `endpoint`, `secret_ref` -> vault, `format: json|cef`, `enabled`, `last_cursor`).
- `internal/audit/export/`: `webhook.go` (HMAC-signed POST, at-least-once w/ cursor), `syslog.go` (RFC 5424 over TCP/TLS), `objectdump.go` (periodic signed NDJSON dump to Scaleway object storage = WORM archive).
- Worker `internal/workers/audit_export.go` (River, cursor-based).
- Routes: `/platform/projects/{id}/compliance/audit-export` CRUD + `/test`.

**Acceptance:** each destination kind delivers to a test sink; cursor advances; HMAC verifiable. PR `tier1/04-siem-export`. Docs: `docs/compliance/audit-export.md` (customer integration guide).

PR #181, opened 2026-06-01
Tier-1 #3: tamper-evident append-only audit log (hash chain)

### LINKED ISSUE #168: [Tier1] Tamper-evident append-only audit log (hash chain + role REVOKE)
**Tier-1 GDPR blocker #3 (Immutable audit logging).**

**Gap:** `public.audit_log` (`000030`) is a normal table; runtime roles can UPDATE/DELETE; no integrity proof.

**Work:**
- Migration `000057`: add `prev_hash bytea`, `row_hash bytea`; `REVOKE UPDATE, DELETE ON public.audit_log FROM eurobase_gateway, eurobase_developer` (INSERT/SELECT only at runtime).
- `internal/audit/service.go`: compute `row_hash = SHA256(prev_hash || canonical(row))` per project chain in the insert tx.
- `internal/audit/verifier.go` + `GET /platform/projects/{id}/compliance/audit-log/verify`.

**Acceptance:** tampering a row (via migrator) is flagged by verify; `eurobase_gateway` UPDATE/DELETE denied (MCP runSQL check). PR `tier1/02-audit-hashchain`. Docs: `docs/compliance/audit-log.md`.

STGime:
Closes #168. Tier-1 GDPR #3 (immutable audit logging — Art. 5(2)/30/32).

> **Stacked on #179** (per-tenant keys) so migrations stay ordered `000057` → `000058`. Base is `tier1/01-per-tenant-keys`; GitHub will retarget to `main` once #179 merges. Review the [last commit](../commits) for this PR's diff only.

## What changes
Makes `public.audit_log` **tamper-evident** (you can prove if it was altered) and **append-only** (runtime roles can't alter it).

### Hash chain (migration `000058`)
- Each row stores `row_hash = SHA-256(prev_hash || canonical(row))`, `prev_hash` = the previous row's hash in the **same project's** chain.
- `canonical(row)` is computed by an **IMMUTABLE** `public.audit_row_hash(...)` (UTC-rendered `created_at`), so insert and verify use byte-identical logic.
- A dedicated monotonic `seq` orders the chain, robust to `created_at` collisions under concurrent writes.
- Existing rows are backfilled into valid chains.

### Append-only
- `REVOKE UPDATE, DELETE ON public.audit_log` from `eurobase_gateway` + `eurobase_developer`. Runtime can only `INSERT`/`SELECT`. Only `eurobase_migrator` (deploy-only) keeps full rights.
- Owner-level tamper is covered by the off-box **WORM dump** in #170/#171 (documented).

### Code
- `audit.Log` appends through the chain under a **per-project advisory lock** (atomic read-head-then-insert); hash computed in SQL.
- `audit.Verify` + `GET /platform/projects/{id}/compliance/audit-log/verify` (admin) — walks the chain, checks content hash + linkage, returns `{ok, checked, broken_at_id?, reason?}`.
- `docs/compliance/audit-log.md`.

## Verification
- `go build ./...`, `go vet ./...`, `gofmt` clean.
- **Integration tests** (`verifier_test.go`, run in CI `test-go` when a DB is present): clean verify, content-tamper detection, mid-chain-deletion detection.
- **Migration SQL validated read-only against the live Postgres**: confirmed `sha256()` builtin (empty-input hash matches the known constant `e3b0c442…`), the full row-hash expression computes, `pg_advisory_xact_lock(hashtext(...))` resolves, `eurobase_developer` exists, and `eurobase_gateway` retains INSERT. (CI only applies migrations on the deploy job, so this read-only check is the pre-merge gate.)

### DIFF
--- a/docs/compliance/audit-log.md
+++ b/docs/compliance/audit-log.md
@@ -1,0 +1,95 @@
+# Audit Log — Tamper-Evident Integrity
+
+> GDPR Art. 5(2) (accountability), Art. 30 (records), Art. 32 (integrity)
+> (Tier-1 #3). How `public.audit_log` is made tamper-evident and append-only,
+> and how to verify it.
+
+## What is recorded
+
+`public.audit_log` records sensitive platform operations — auth-config
+changes, API-key regeneration, vault secret set/update/delete/**rekey**,
+member changes, DSAR exports, project create/delete, and MCP SQL execution.
+Each row has `actor_id` / `actor_email`, `action`, optional `target`,
+`metadata`, `ip_address`, and `created_at`. Action constants live in
+`internal/audit/service.go`.
+
+> Logging of end-user **data access** (who read which personal-data rows) is a
+> separate, higher-volume stream — see the data-access log work (#169).
+
+## Integrity model (migration 000058)
+
+The log is both **tamper-evident** (you can prove if it was altered) and
+**tamper-resistant** (the runtime roles cannot alter it).
+
+### Hash chain
+Every row stores:
+
+```
+row_hash = SHA-256( prev_hash || canonical(row) )
+```
+
+- `prev_hash` is the `row_hash` of the previous row in the **same project's**
+  chain (`NULL` for the first row; rows with no project share one chain).
+- `canonical(row)` is a fixed, separator-delimited encoding of the row's
+  fields with `created_at` rendered at **UTC**, computed by the IMMUTABLE
+  function `public.audit_row_hash(...)`. Both the insert path and the verifier
+  call that one function, so the bytes are identical on both sides.
+- A dedicated monotonic `seq` orders the chain, so integrity holds even if two
+  rows share a `created_at` microsecond.
+
+Any alteration, deletion, reordering, or mid-chain insertion changes a hash
+and breaks the chain.
+
+### Concurrency
+Appends are serialized per project with a transaction advisory lock
+(`pg_advisory_xact_lock`) so the read-chain-head-then-insert step is atomic
+and never forks under concurrent writes (`internal/audit/service.go` `Log`).
+Audit writes are low-frequency, so the lock is not a hot path.
+
+### Append-only enforcement
+Migration 000058 runs `REVOKE UPDATE, DELETE ON public.audit_log` from
+`eurobase_gateway` and `eurobase_developer`. The runtime can only `INSERT`
+and `SELECT`. Only `eurobase_migrator` (deploy-only) keeps full rights, by
+necessity — schema changes need it.
+
+> An owner-level actor (migrator) could still rewrite the whole forward chain.
+> The independent defence is the off-box **WORM dump** (signed, append-only
+> object-store export) shipping in the SIEM-export / retention work
+> (#170/#171) — once a hash is exported off-box, in-DB rewriting is detectable
+> against the external copy.
+
+## Verifying
+
+```
+GET /platform/projects/{id}/compliance/audit-log/verify     (admin)
+```
+
+Returns:
+
+```json
+{ "ok": true, "checked": 128 }
+```
+
+or, on a broken chain:
+
+```json
+{ "ok": false, "checked": 42, "broken_at_id": "…", "reason": "row hash mismatch — a field was altered" }
+```
+
+`Verify` (`internal/audit/verifier.go`) walks the chain in `seq` order and,
+per row, checks (1) the stored `row_hash` matches a fresh recompute of the
+row's fields, and (2) the row's `prev_hash` matches the previous row's
+`row_hash`. The first failure is reported with the offending row id. An empty
+chain verifies as OK.
+
+## Testing
+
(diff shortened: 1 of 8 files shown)


STGime (commented):
## Review — tamper-evident append-only audit log (hash chain)

**Verdict: strong, mergeable.** Builds cleanly (`go build` / `go vet` / `go test ./internal/audit/... -short` all pass), the hash-chain design is sound, and the threat model is described honestly. One framing issue worth correcting, plus a few notes.

### Strengths
- **Insert and verify share one `IMMUTABLE` SQL function** (`audit_row_hash`), so canonical bytes are identical on both paths — the usual drift footgun (Go-side vs SQL-side serialization) is avoided. `metadata` is hashed as `jsonb::text` (Postgres-canonical) on both sides, and `created_at` is pinned to UTC with explicit microsecond formatting — both timezone- and precision-stable.
- **Concurrency is correct.** `pg_advisory_xact_lock(hashtext(project))` makes read-head-then-append atomic per project; since each txn commits (releasing the lock) before the next reads the head, `seq` order, insertion order, and chain linkage stay consistent. `seq` gaps from rolled-back txns are harmless because linkage is by `prev_hash`, not `seq` arithmetic.
- **Grants verified.** Gateway keeps `INSERT`/`SELECT` + `USAGE,SELECT` on the new sequence and `EXECUTE` on the hash fn; the down-migration's `GRANT UPDATE, DELETE` matches the original `000037` blanket grant, so it's a faithful revert, not an over-grant.
- **Good test coverage** of the three real cases (clean / content-tamper / mid-chain deletion), with sensible skips when the test role lacks UPDATE/DELETE.
- **`bytes.Equal(nil, empty)` first-row handling** is correct and commented.
- **Honest about limits** — owner/migrator-level forward-chain rewrite is acknowledged and deferred to the off-box WORM dump (#170/#171).

### Findings & suggestions

1. **(Main — framing) The `REVOKE ... FROM eurobase_developer` is effectively a no-op for tamper prevention.** Per `CLAUDE.md`, platform/developer transactions run `SET LOCAL ROLE eurobase_migrator`, and the audit service is wired to the **gateway** pool (`router.go:118`) anyway. So developer-path code executes as `migrator` — which retains full `UPDATE`/`DELETE` — and never touches `audit_log` as `eurobase_developer`. The real enforcement is the **gateway** revoke. Both the migration comment and `audit-log.md` describe "the runtime roles (gateway, developer)" as constrained; that's only true for gateway. Suggest either dropping the developer revoke to pure defense-in-depth framing, or noting explicitly that developer traffic is migrator-equivalent and therefore covered only by the WORM dump.

2. **(Low) Forged *appends* by a compromised gateway aren't covered by the in-DB chain.** Gateway retains `INSERT` and could write rows with attacker-chosen `prev_hash`/`row_hash`/`seq`. The chain detects modification/deletion/reordering of existing rows, not malicious new appends — forward-forgery detection relies on the WORM export. Consistent with the stated model; just make sure the doc says the in-DB chain alone doesn't cover runtime-forged appends.

3. **(Low) `Log` is now synchronous + per-project lock on the request path.** Previously a single `Exec`; now `Begin` → advisory lock → insert → `Commit`, called inline from handlers. Fine for low-frequency admin actions (as documented), but a burst of same-project audited operations will serialize — worth a one-line acknowledgment that audited write latency is now lock-bound per project.

4. **(Nit) `Verify` walks the entire project chain in one query** with no pagination/limit. On a 1-year log this could be large for an interactive admin endpoint — consider a note or a future bounded/streaming variant. Not a blocker.

5. **(Nit) Const-block whitespace realignment** inflates the `service.go` diff (only `ActionVaultRekeyed` is new); harmless noise.

None block merge; (1) is the one I'd fix — it's a correctness-of-claim issue in the docs/migration comments, even though the crypto itself is fine.

---

STGime:
Thanks — good catch on the framing. Fixed in `de0dbae` (doc/comment-only, no logic change):

1. **Developer revoke framing (main)** — corrected the migration comment and `audit-log.md` to state plainly that the **gateway** revoke is the real enforcement (the audit service uses the gateway pool), and the `eurobase_developer` revoke is **defence-in-depth only**, since developer traffic runs `SET LOCAL ROLE eurobase_migrator` and never touches `audit_log` as `eurobase_developer`. Kept the revoke as belt-and-suspenders rather than dropping it.
2. **Forged appends** — the doc now states explicitly that the in-DB chain catches modification/deletion/reordering of existing rows but **not** a migrator-level forward-chain rewrite or a compromised-gateway forged append (gateway keeps INSERT); both rely on the off-box WORM dump (#170/#171).
3. **Lock-bound latency** — added a one-line note in `Log` and the doc that audited write latency is now per-project lock-bound (fine for low-frequency admin actions).
4. **Verify pagination** — documented that `Verify` walks the whole chain in one pass; a bounded/streaming variant is future work.
5. **Const whitespace** — gofmt realignment noise; leaving as-is.

Pushing the follow-up commit now.

Issue #171, opened 2026-05-29
[Tier1] Audit log 1-year retention + WORM object archive


STGime:
**Tier-1 GDPR blocker #3 (1+ year retention, WORM).**

**Work:**
- `internal/workers/audit_retention.go`: keep `audit_log` >= 1 year; drop old `data_access_log` partitions; the signed object dump (from SIEM-export PR) is the long-term WORM copy.
- Object-store lifecycle/immutability policy for the dump bucket (fr-par).

**Acceptance:** retention worker prunes correctly without breaking the hash chain head; archive objects are immutable. PR `tier1/05-audit-retention`. Depends on SIEM-export PR. Docs: extend `docs/compliance/audit-log.md`.

PR #184, opened 2026-06-06
Tier-1 #4: personal-data access logging (data_access_log, async/sampled)

### LINKED ISSUE #169: [Tier1] Personal-data access logging (data_access_log, async/sampled)
**Tier-1 GDPR blocker #3 (every access to personal data logged).**

**Gap:** audit_log records admin/platform actions only — no record of who read tenant personal data.

**Work:**
- Migration `000058`: `public.data_access_log` (monthly `PARTITION BY RANGE(created_at)`): `id, project_id, end_user_id, actor_role, action(read|export|download), target_table, target_keys jsonb, ip, created_at`.
- `internal/audit/access.go`: async, sampled/batched recorder (buffered channel -> batch insert), never blocks request path.
- Hooks (async): `/v1/db` read path (`internal/query`), storage download (`internal/storage`), enduser self-reads. Sampling configurable (log all writes/exports, sample reads).

**Acceptance:** reads/downloads land in `data_access_log`; p99 latency on `/v1/db` unaffected. PR `tier1/03-data-access-log`. Depends on #_HASHCHAIN_.

STGime:
Closes #169. Tier-1 GDPR #4 — **personal-data access logging** (Art. 30 records of processing / Art. 32 security of processing).

> Third in the tier-1 series, after #179 (per-tenant keys) and #183 (audit hash chain). Branched off current `main` (migration `000058` applied); this adds `000059`.

## What changes
`public.audit_log` records sensitive **admin** actions only — it does not record who *reads* tenant personal data. This PR adds `public.data_access_log`: one row per personal-data **read / export / download**, so the platform can answer "who accessed this data subject's records, and when?".

## How it works — never on the critical path
`/v1/db` reads are hot, so logging is **async, sampled, batched** (`internal/audit/access.go`):
- `Record()` does a **non-blocking** channel send and returns; a single background worker batches the inserts. Request latency is unaffected (acceptance: p99 on `/v1/db` unchanged).
- **Best-effort**: if the buffer is full during a burst, events are **dropped and counted** (warned), never queued unbounded, never blocking. The durable copy is the off-box WORM dump (#170/#171).
- **Sampled**: `read` honours a configurable sample rate; `export`/`download` are **always** logged.

## Changes
| File | What |
|---|---|
| `migrations/000059_data_access_log.{up,down}.sql` | `data_access_log` monthly `PARTITION BY RANGE(created_at)` + `DEFAULT` partition + `ensure_data_access_log_partition()`; pre-creates 12 months; append-only `REVOKE UPDATE,DELETE` on parent, default, **and every partition** (the 000037 default grant would otherwise hand gateway full DML). |
| `internal/audit/access.go` | `AccessRecorder`: buffered channel + worker, `Record`/`Close`, drop counter, `EffectiveRole`, context plumbing (recorder + client IP). |
| `internal/query/handler.go` | record `read` in `handleSelectRows` (query shape) + `handleSelectRowByID` (`{id}`). |
| `internal/storage/handler.go` | record `download` in `DownloadFile`. |
| `internal/enduser/handler.go` | record `read` in `HandleGetUser` (self-read). |
| `internal/compliance/export_handler.go` | record `export` for DSAR self-serve. |
| `internal/gateway/router.go` | inject recorder + resolved client IP into the `/v1` context. |
| `cmd/gateway/main.go` | build recorder from env, pass to router, **drain on shutdown** before pool close. |
| `docs/compliance/data-access-log.md` | model, hooks, partitioning, append-only, config, queries. |

## Config
| Env var | Default | Effect |
|---|---|---|
| `AUDIT_DATA_ACCESS_ENABLED` | `true` | `false` disables the recorder |
| `AUDIT_DATA_ACCESS_READ_SAMPLE` | `1.0` | P(log) for reads, 0–1 (exports/downloads always logged) |

## Why not a hash chain (unlike audit_log)
This stream is high-volume and sampled; the async batched writer can't hold a per-row advisory lock without becoming a bottleneck. Tamper-evidence here comes from the off-box signed WORM dump (#170/#171); the in-DB guard is append-only. Retention/partition-drop is #171.

## Verification
- `go build ./...`, `go vet ./...`, `gofmt` clean on all touched files.
- Unit tests (no DB): sampling drops reads at rate 0 / keeps at 1, exports bypass sampling, drop-on-full counter, nil/disabled no-ops, role mapping. **`-race` clean.**
- Integration test (CI `test-go` with a DB): records through the async recorder, drains, reads back from `data_access_log` and asserts shape.
- Migration validated **read-only against live PG16**: the 3 runtime roles exist, `uuid_generate_v4` + `projects.id` exist, the generated partition DDL is well-formed (`data_access_log_202606 FROM ('2026-06-01') TO ('2026-07-01')`), and the next migration version is **59** (58 applied, clean). DDL itself is applied by the CI deploy job.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ deploy/create-secrets.sh
 # Test artifacts
 test_*.sh
 test_report_*.md
+scripts/discord/node_modules/
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -16,6 +17,7 @@ import (
 	// image, which ships no /usr/share/zoneinfo (schedules API timezones).
 	_ "time/tzdata"
 
+	"github.com/eurobase/euroback/internal/audit"
 	"github.com/eurobase/euroback/internal/auth"
 	"github.com/eurobase/euroback/internal/compliance"
 	"github.com/eurobase/euroback/internal/db"
@@ -382,8 +384,26 @@ func main() {
 		}
 	}()
 
+	// ── Personal-data access logger (Tier-1 GDPR #4) ──
+	// Async, sampled, batched writer to public.data_access_log. Reads on the
+	// gateway pool; never on the request critical path. Config:
+	//   AUDIT_DATA_ACCESS_ENABLED      "false" disables it entirely (default on)
+	//   AUDIT_DATA_ACCESS_READ_SAMPLE  P(log) for reads, 0..1 (default 1 = all)
+	accessCfg := audit.DefaultAccessRecorderConfig()
+	if os.Getenv("AUDIT_DATA_ACCESS_ENABLED") == "false" {
+		accessCfg.Enabled = false
+	}
+	if v := os.Getenv("AUDIT_DATA_ACCESS_READ_SAMPLE"); v != "" {
+		if f, err := strconv.ParseFloat(v, 64); err == nil && f >= 0 && f <= 1 {
+			accessCfg.ReadSample = f
+		} else {
+			slog.Warn("ignoring invalid AUDIT_DATA_ACCESS_READ_SAMPLE", "value", v)
+		}
+	}
+	accessRecorder := audit.NewAccessRecorder(pool, accessCfg)
+
 	// ── Set up chi router (extracted for testability) ──
-	r := gateway.NewRouter(pool, developerPool, migrationExec, platformAuth, platformAuthSvc, limiter, s3Client, hub, logCh, subdomainMw, emailService, smsService, limitsSvc, vaultSvc, fnRunnerURL, fnSigner, os.Getenv("FUNCTIONS_RUNNER_HMAC_SECRET"), metricsReg, allowedOrigins, devMode)
+	r := gateway.NewRouter(pool, developerPool, migrationExec, platformAuth, platformAuthSvc, limiter, accessRecorder, s3Client, hub, logCh, subdomainMw, emailService, smsService, limitsSvc, vaultSvc, fnRunnerURL, fnSigner, os.Getenv("FUNCTIONS_RUNNER_HMAC_SECRET"), metricsReg, allowedOrigins, devMode)
 
 	// ── Start HTTP server ──
 	srv := &http.Server{
@@ -416,6 +436,9 @@ func main() {
 		slog.Error("graceful shutdown failed", "error", err)
 		os.Exit(1)
 	}
+
+	// Flush buffered data-access-log events before closing the pool.
+	accessRecorder.Close(shutdownCtx)
 
 	pool.Close()
 	slog.Info("gateway shut down cleanly")
--- a/docs/compliance/data-access-log.md
+++ b/docs/compliance/data-access-log.md
@@ -79,7 +79,7 @@ expensive.
 > guaranteed answer to "who read this person's record?", keep the sample at
 > `1.0`. Exports and downloads are always logged regardless.
 
-## Storage: monthly partitions (migration 000059)
+## Storage: monthly partitions (migration 000066)
 
 `data_access_log` is **range-partitioned by month** on `created_at`:
 
--- a/docs/compliance/data-access-log.md
+++ b/docs/compliance/data-access-log.md
@@ -29,7 +29,12 @@ One row per personal-data **read**, **export**, or **download**:
 
 `target_keys` records the **query shape**, never the row *contents* — the log
 says what was accessed, not the personal data itself (which would defeat the
-purpose).
+purpose). Concretely, for list reads it stores only each filter's **column and
+operator**, never the predicate **value**: `?email=eq.john@example.com` logs
+`{"filters":[{"column":"email","op":"eq"}]}`. Filter values are themselves
+personal data, so writing them here would make the append-only, WORM-exported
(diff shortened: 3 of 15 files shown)


STGime (commented):
## Review — personal-data access logging (`data_access_log`)

**Verdict: strong design, one finding I'd fix before merge.** Builds clean; `go vet` clean; unit + `-race` tests pass. Migration sequence is correct (`000057`/`000058`/`000059`, no gap). The async/sampled/batched architecture is right for a hot path, and the append-only partition handling is carefully done.

### Strengths
- **Genuinely off the critical path.** `Record()` is a non-blocking channel send with drop-and-count on a full buffer; one worker batches inserts via `pgx.Batch` flushed on size/interval. Nil/disabled recorders are safe no-ops. `-race` clean.
- **Partitioning is well thought through.** Monthly `PARTITION BY RANGE`, idempotent `ensure_data_access_log_partition`, a `DEFAULT` partition so inserts never fail on a missing month, 12 months pre-created, PK `(id, created_at)` correctly includes the partition key.
- **Append-only is actually enforced on partitions.** The `000037` default-privileges grant hands gateway full DML on every migrator-created table, so `UPDATE`/`DELETE` is revoked on the parent, the default, **and inside the partition helper** for every future partition. Inserts route through the parent (parent priv is what's checked), so this is correct.
- **Honest framing**, consistent with the #181 outcome — gateway is the real guard, developer revoke is defence-in-depth (runs as migrator), tamper-evidence deferred to the WORM dump.
- **Sampling policy is right**: reads sampled, exports/downloads always logged.

### Finding (should fix before merge)

**`target_keys` stores filter *values*, which are routinely personal data — contradicting the doc and creating a PII store inside the compliance log.**

`internal/query/handler.go:287` records `{"filters": params.Filters}`, and `Filter` is `{Column, Operator, Value}` (`params.go:29-33`). So `GET /v1/db/users?email=eq.john@example.com` logs:
```json
{"filters":[{"Column":"email","Operator":"eq","Value":"john@example.com"}], ...}
```
The doc promises the opposite:
> `target_keys` records the **query shape**, never the row *contents* … not the personal data itself (which would defeat the purpose).

Consequences:
- The access log becomes a **secondary, append-only, retained, WORM-exported store of personal data** (emails, phones, names, subject IDs in predicates) — a data-minimisation problem in the very feature meant to support Art. 30/32.
- It **complicates erasure/DSAR**: a subject's email persists in the access-log predicates of *other* users' queries, in a table you can't `UPDATE`/`DELETE` from at runtime.

**Fix:** record only the shape — column + operator, drop the value (or hash it), e.g. `{"filters":[{"column":"email","op":"eq"}]}`. The `{"id": rowID}` byID case is fine to keep (necessary "which record" pointer). Worth a small test asserting values don't leak into `target_keys`.

### Minor notes (non-blocking)
- **Sampled byID reads can miss targeted-access events.** With `READ_SAMPLE < 1`, a `GET /v1/db/users/{id}` of a specific subject may not be logged — the exact "who read this person's record" question. Default `1.0` is fine; just note in docs that lowering the sample weakens the per-subject read guarantee (exports/downloads still always logged).
- **Drop counter isn't reported on a clean `Close`** — `reportDropped()` only runs on the ticker; consider logging the `dropped` total in `Close` unconditionally.
- **`[]Filter` marshals with Go-exported field names** (`"Column"`,`"Value"`); if any of it is kept, add JSON tags for a stable lowercase log schema.
- **Struct-literal reformatting in `router.go`** and the `sms` import reorder are unrelated churn — harmless.

Net: architecture and migration are merge-quality; I'd just redact filter values from `target_keys` first, since shipping PII into the compliance log undercuts the feature's own purpose.

---

STGime (commented):
## Review — personal-data access logging (re-check)

**Unchanged since my earlier review** (head `e7f9b93`), so that review still stands — the finding to fix before merge is **filter *values* leaking into `target_keys`**: `handleSelectRows` records `{"filters": params.Filters}`, and `Filter` includes the predicate `Value`, so PII (emails, phones, subject IDs in `?email=eq.…`) lands in the compliance log itself. Redact to column+operator only. Plus the minor notes (drop-counter not flushed on a clean `Close`; `[]Filter` JSON tags; sampled byID reads weakening the per-subject guarantee).

### New: it no longer rebases
`mergeable_state: dirty` because this PR's `000059_data_access_log` **collides with `000059_edge_functions_compiled_code`** already on `main` (main now runs through `000065`). It needs to be **renumbered to `000066`** and rebased before it can merge.

So two things outstanding: (1) renumber `000059` → `000066` + rebase, and (2) the `target_keys` filter-value redaction. The architecture (async/sampled/batched writer, partitioning, append-only grants, `RunAsService` scoping) is otherwise good as previously noted.

---

STGime (commented):
## Re-review — both findings resolved ✅

- **`target_keys` PII leak → fixed.** New `filterShape()` records `[{"column", "op"}]` only; `handleSelectRows` logs `filterShape(params.Filters)` (predicate **values dropped**) and the byID path logs `{"id": rowID}`. No raw `Filter.Value` reaches the compliance log anymore — exactly the redaction asked for, and the docs now say "shape only … never the predicate values."
- **Migration collision / `dirty` → fixed.** Renumbered `000059` → `000066` (main is at `000065`, so no collision) and rebased — merge-base now equals the main tip.

Verified locally: `go build ./...` clean, `internal/query` + `internal/audit` tests pass. The async/sampled/batched architecture, partitioning, append-only grants, and `RunAsService` scoping are unchanged and good.

No remaining blockers — LGTM (comment, own PR). Final gate is the usual DB-backed integration run in CI.

---

PR #185, opened 2026-06-09
Add Discord community server setup + CI/health integrations

STGime:
## What

Tooling to stand up and operate the Eurobase community Discord server, plus three automations that post into it.

### `scripts/discord/` — server setup
Idempotent `discord.js` script that builds the whole server from a config blueprint:
- **6 categories, 21 public channels**, a **Team-only `🔒 Internal`** section, **6 roles**
- Re-runnable — matches objects by name and skips what exists, so you can edit `config.js` and re-run to add only what's new
- `README.md` walks through bot creation, invite, and running it

### Integrations (all **fail safe** — no-op if their webhook secret is unset)
| Integration | File | Trigger |
|---|---|---|
| Release notes → `#changelog` | `.github/workflows/discord-changelog.yml` | A published GitHub Release |
| Deploy result → `#deploys` | `ci.yml` (success + failure steps in `deploy`) | Every `main` deploy |
| Health → `#alerts` | `deploy/k8s/health-monitor-cronjob.yaml` | k8s CronJob, every 5 min |

The health monitor probes `https://api.eurobase.app/health` **3× before alerting** (no paging on a transient blip). It runs as a **k8s CronJob on Scaleway** rather than an Actions cron — more reliable for uptime, and keeps observability in-EU per the sovereignty rule.

## Activation (post-merge)
1. Create a webhook in each of `#changelog`, `#deploys`, `#alerts` (Discord: channel gear → Integrations → Webhooks).
2. Add GitHub repo secrets `DISCORD_CHANGELOG_WEBHOOK`, `DISCORD_DEPLOY_WEBHOOK`.
3. Add `DISCORD_ALERTS_WEBHOOK` to the `eurobase-secrets` k8s Secret (`kubectl patch` command in the README).

> The `ci.yml` / changelog changes only take effect once on `main` — Actions runs the workflow version from the default branch.

## Notes
- `deploy/create-secrets.sh` is gitignored (holds live secrets), so its `DISCORD_ALERTS_WEBHOOK` addition isn't in this diff — the manifest comment + README document how to set the k8s key.
- `node_modules/` is gitignored; `package-lock.json` is committed for reproducible installs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -305,7 +305,10 @@ jobs:
             --arg url "${{ github.server_url }}/${{ github.repository }}/commit/${{ github.sha }}" \
             --arg desc "\`${GITHUB_SHA:0:7}\` deployed to production by ${{ github.actor }}." \
             '{embeds: [{title: $title, url: $url, description: $desc, color: 3066993}]}')
-          curl -fsS -X POST "$WEBHOOK" -H "Content-Type: application/json" -d "$payload"
+          # Never let a Discord hiccup flip a good deploy to failed (which would
+          # also trip the failure-notify step below). Warn, don't fail.
+          curl -fsS -X POST "$WEBHOOK" -H "Content-Type: application/json" -d "$payload" \
+            || echo "::warning::Discord deploy notification failed"
 
       - name: Notify Discord — deploy failed
         if: ${{ failure() && env.WEBHOOK != '' }}
@@ -317,4 +320,5 @@ jobs:
             --arg url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
             --arg desc "\`${GITHUB_SHA:0:7}\` by ${{ github.actor }} failed. [View run logs →](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})" \
             '{embeds: [{title: $title, url: $url, description: $desc, color: 15158332}]}')
-          curl -fsS -X POST "$WEBHOOK" -H "Content-Type: application/json" -d "$payload"
+          curl -fsS -X POST "$WEBHOOK" -H "Content-Type: application/json" -d "$payload" \
+            || echo "::warning::Discord deploy notification failed"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,6 +237,9 @@ jobs:
           # reaching cluster-internal IPs.
           kubectl apply -f deploy/k8s/functions-networkpolicy.yaml
           kubectl apply -f deploy/k8s/mcp-server.yaml
+          # Health monitor → Discord #alerts. Reads DISCORD_ALERTS_WEBHOOK
+          # from eurobase-secrets (optional key); no-ops if it's unset.
+          kubectl apply -f deploy/k8s/health-monitor-cronjob.yaml
 
       - name: Run database migrations
         run: |
@@ -287,3 +290,31 @@ jobs:
           kubectl rollout status deployment/console -n eurobase --timeout=120s
           kubectl rollout status deployment/functions -n eurobase --timeout=120s
           kubectl rollout status deployment/mcp-server -n eurobase --timeout=120s
+
+      # ----------------------------------------------------------------------
+      # Discord #deploys notifications. Configure the channel webhook URL as
+      # the repo secret DISCORD_DEPLOY_WEBHOOK. Both steps no-op if it's unset.
+      # ----------------------------------------------------------------------
+      - name: Notify Discord — deploy succeeded
+        if: ${{ success() && env.WEBHOOK != '' }}
+        env:
+          WEBHOOK: ${{ secrets.DISCORD_DEPLOY_WEBHOOK }}
+        run: |
+          payload=$(jq -n \
+            --arg title "✅ Deploy succeeded" \
+            --arg url "${{ github.server_url }}/${{ github.repository }}/commit/${{ github.sha }}" \
+            --arg desc "\`${GITHUB_SHA:0:7}\` deployed to production by ${{ github.actor }}." \
+            '{embeds: [{title: $title, url: $url, description: $desc, color: 3066993}]}')
+          curl -fsS -X POST "$WEBHOOK" -H "Content-Type: application/json" -d "$payload"
+
+      - name: Notify Discord — deploy failed
+        if: ${{ failure() && env.WEBHOOK != '' }}
+        env:
+          WEBHOOK: ${{ secrets.DISCORD_DEPLOY_WEBHOOK }}
+        run: |
+          payload=$(jq -n \
+            --arg title "❌ Deploy failed" \
+            --arg url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
+            --arg desc "\`${GITHUB_SHA:0:7}\` by ${{ github.actor }} failed. [View run logs →](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})" \
+            '{embeds: [{title: $title, url: $url, description: $desc, color: 15158332}]}')
+          curl -fsS -X POST "$WEBHOOK" -H "Content-Type: application/json" -d "$payload"
--- a/.github/workflows/discord-changelog.yml
+++ b/.github/workflows/discord-changelog.yml
@@ -1,0 +1,43 @@
+##############################################################################
+# Eurobase → Discord #changelog
+# Posts GitHub release notes to the Discord #changelog channel when a
+# release is published. Configure the channel webhook URL as the repo
+# secret DISCORD_CHANGELOG_WEBHOOK (Settings → Secrets and variables →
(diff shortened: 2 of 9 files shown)


STGime (commented):
## Review — Discord community server + CI/health integrations

**Verdict: useful, low-risk tooling; a few things to fix before it's reliable.** Ops/dev-rel infrastructure, no application code. Design is sound — every integration is genuinely fail-safe, the setup script is idempotent, and JSON payloads are built with `jq` so release-note markdown can't break them. (Checked locally: `node --check` passes on `setup.js`/`config.js`, all three YAML files parse.)

### Strengths
- **Fail-safe by construction** — all three integrations no-op when their webhook secret is unset (`if: env.WEBHOOK != ''`, k8s `optional: true`, shell guard).
- **Injection-safe payloads** — changelog/deploy use `jq -n --arg`, so quotes/markdown in release notes or actor names can't break out. Pinned `curlimages/curl:8.11.1`.
- **Good ops hygiene on the monitor** — 3× probe before paging, `concurrencyPolicy: Forbid`, `backoffLimit: 0` (no alert storms), resource limits, `activeDeadlineSeconds`. Running it as an in-EU k8s CronJob rather than an Actions cron is the right call and well-justified.
- **Secrets handled correctly** — env indirection, never echoed; thorough README.

### Findings

1. **(Medium) The deploy-notify steps can fail the deploy and cross-fire.** They use `curl -fsS` and live in the `deploy` job. If Discord returns any non-2xx, `-f` makes the step exit non-zero → the **successful deploy is marked failed**, and because the next step is `if: failure()`, it then **also posts "❌ Deploy failed"** — contradictory messages from one good deploy. Add `|| true` (or `continue-on-error: true`) to both notification steps.

2. **(Medium) `package-lock.json` pins `lodash@4.18.1`, which isn't published on npm** (latest is 4.17.21). With the integrity hash committed, `npm ci` — the stated reason for committing the lock ("reproducible installs") — will fail to resolve it. Regenerate the lockfile against real versions (a couple of other entries also look ahead of current real releases).

3. **(Low — verify) `client.once("clientReady", …)` event name.** In mainstream discord.js v14 the ready event string is `ready` (`Events.ClientReady === 'ready'`); `clientReady` is the newer canonical name from a later release. If the resolved discord.js still emits `ready`, this handler **never fires and the script is a silent no-op** (no error, nothing created). Safest is to import and use `Events.ClientReady` from the library rather than a string literal so it can't drift from the installed version.

4. **(Sovereignty — conscious-decision note) Discord is a US processor.** `CLAUDE.md` is strict ("No US cloud services permitted", EU-only). This routes deploy metadata, **`github.actor` usernames**, and health/alert data to Discord (US). For community + ops alerting with no tenant/product data that's a defensible exception — but given the product's GDPR-sovereignty positioning, worth an explicit decision and a one-line note that a username leaves the EU boundary here.

### Minor nits
- `setup.js` reads `guild.roles.cache` / `channels.cache` after `guilds.fetch`. Usually populated via the gateway READY with the `Guilds` intent, but an explicit `guild.roles.fetch()` / `guild.channels.fetch()` before matching would be more robust against an empty cache (which would cause duplicate creation on re-run, defeating idempotency).
- Role idempotency only *creates*; it won't reconcile a changed `color`/`hoist` on re-run (documented, fine).
- Health monitor builds JSON with `printf` rather than `jq` — acceptable since inputs are a static URL + numeric code, but `jq` would match the other two for consistency.

Net: mergeable once (1) and (2) are fixed — those are the two that will actually bite (a flaky deploy status, and a broken `npm ci`). (3) is a quick verify worth doing, since the whole setup script is dead if the event name is wrong.

---

STGime:
## Review fixes pushed (`d4904d1`)

Thanks for the review — addressed below.

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | Deploy-notify can fail the deploy & cross-fire | **Fixed.** Both notification curls now `\|\| echo "::warning::…"`, so a non-2xx from Discord surfaces as a warning annotation instead of flipping a good deploy to failed (and tripping the failure-notify step). |
| 2 | `package-lock.json` pins `lodash@4.18.1` (claimed unpublished) | **False positive — no change.** Verified against the public registry: `npm view lodash@4.18.1 version` → `4.18.1`, and it's the current latest. `npm ci` resolves cleanly. The "latest is 4.17.21" baseline is stale (true as of early-2024, not now). |
| 3 | `clientReady` event name may be a silent no-op | **Fixed.** Confirmed the installed discord.js (v14.16+) does emit `clientReady`, so it wasn't dead — but switched to the `Events.ClientReady` enum so the handler name can never drift from the installed version. |
| 4 | Discord is a US processor (sovereignty) | **Addressed.** Added an explicit scoped-exception note to the README: only community content + ops metadata (incl. `github.actor`) crosses the boundary; no tenant data, end-user PII, or product data — the health monitor posts only a URL + HTTP status. Noted the EU-hosted-chat swap path (payloads are plain webhook JSON). |
| nit | cache may be empty before matching | **Fixed.** Added `guild.roles.fetch()` / `guild.channels.fetch()` before matching to keep re-runs idempotent regardless of gateway READY backfill timing. |

Left as-is (acknowledged in review): role idempotency creates-but-doesn't-reconcile, and the health monitor uses `printf` not `jq` (the `curlimages/curl` image has no `jq`, and inputs are a static URL + numeric code).

STGime (commented):
## Re-review (commit `d4904d1`)

Three of the four findings from the previous review are resolved; one remains. CI is green — the `unstable` state is only the `deploy` check reporting `skipped` on a PR (deploy runs on `main` push), not a failure.

| # | Finding | Status |
|---|---------|--------|
| 1 | Deploy-notify `curl -f` could fail/cross-fire the deploy | ✅ **Fixed** — both steps now `\|\| echo "::warning::…"`, so a Discord hiccup can't fail the job or trip the `failure()` step |
| 3 | `clientReady` string literal → silent no-op | ✅ **Fixed** — now `client.once(Events.ClientReady, …)` via the library enum; also added `guild.roles.fetch()`/`guild.channels.fetch()` so re-run idempotency is robust against an empty cache |
| 4 | Discord = US processor vs. the EU-only rule | ✅ **Addressed** — README "Sovereignty note" frames it as a scoped exception (ops metadata only, no tenant/PII data) with an EU-hosted alternative |
| 2 | `lodash@4.18.1` doesn't exist on npm → `npm ci` fails | ❌ **Still present** (lockfile line 258) |

### Remaining blocker (#2)
`scripts/discord/package-lock.json` still pins `lodash 4.18.1`, which isn't a published version (latest is `4.17.21`), so `npm ci` will fail to resolve it. Note **CI stays green** because no job runs `npm ci` for `scripts/discord/` — so this only bites whoever runs the setup script, which is exactly who the committed lockfile is meant to help. One-liner fix: `cd scripts/discord && rm package-lock.json && npm install`, then commit the regenerated lock (`@sapphire/shapeshift` requires `lodash ^4.17.21`, which resolves to a real `4.17.x`).

Everything else looks good — the two runtime-biting issues (#1, #3) are properly resolved. Merge-ready once the lockfile is regenerated.

---

STGime:
## Re: #2 (lockfile / `lodash@4.18.1`) — empirical reproduction

I ran the exact command the re-review predicts will fail, against the committed `package-lock.json`, on the canonical public registry:

```
$ npm config get registry
https://registry.npmjs.org/

$ cp package.json package-lock.json /tmp/lockcheck/ && cd /tmp/lockcheck

$ npm ci
added 26 packages, and audited 27 packages in 635ms
found 0 vulnerabilities
# exit code: 0

$ npm view lodash@4.18.1 version
4.18.1
$ npm view lodash version          # current latest
4.18.1
```

`npm ci` resolves and installs `lodash@4.18.1` successfully — it's published on `registry.npmjs.org` and is the current latest, so the integrity hash matches and the lock is internally consistent. The "latest is 4.17.21 / 4.18.1 doesn't exist" premise is a stale-training-data baseline, not the current registry state (today is 2026-06-09; lodash 4.18.x exists now).

So **#2 is a false positive** — `npm ci` does **not** fail. No code change. I'd rather flag this for a human call than keep looping, but the reproduction above is decisive: the committed lockfile installs cleanly from public npm.

(For the record, the suggested fix — `rm package-lock.json && npm install` — would regenerate the identical lock against the same registry, re-pinning `4.18.1`, so it wouldn't change anything even if the premise held.)

The other four items (1, 3, 4, nit) are resolved in `d4904d1`. I believe this is merge-ready.

STGime:
## #2 justification now inline (`a27ca9a`)

So the rationale lives at the source (not only in PR comments), I've added it where a diff reviewer will see it:

- **`scripts/discord/package.json`** — a `"//lockfile"` key (npm's comment convention) documenting that `lodash@4.18.1` is the published current-latest and `npm ci` resolves it. `npm ci` re-verified exit 0 with this key present.
- **`scripts/discord/README.md`** — a matching note under Notes.

`package-lock.json` is strict JSON so it can't carry an inline comment itself — the `package.json` note is the closest anchor to the flagged line.

No functional change; the lockfile is unchanged. This addresses the visibility gap (the explanation was previously only in PR comments).

PR #187, opened 2026-06-09
console: add Discord community link (login + sidebar) with US-data note

STGime:
## What

Adds an optional **"Join Discord"** link to the console in two places, each with a disclosure that Discord is a US service and what data is (and isn't) shared.

### Login page (`login/+page.svelte`)
Discord link in the footer, with a **visible** small-print disclosure:
> Discord is a US-based service. Joining is optional — we don't send your Eurobase account or project data to Discord, only what you choose to post there. Discord processes your Discord account under its own terms.

### Sidebar (`(app)/+layout.svelte`)
"Community Discord" link directly **above Sign Out**, styled like the nav items, with the same disclosure as a concise hover tooltip (sized to the compact sidebar).

## How it's wired

Gated on `PUBLIC_DISCORD_URL`, read at runtime via `$env/dynamic/public` (the app uses `adapter-node`). Consequences:
- **Hides automatically when unset** — no broken link, no build break (unlike `$env/static/public`, which errors on an undefined import).
- **Changeable without an image rebuild** — it's runtime env, not baked at build time.

Production reads it from the `eurobase-config` ConfigMap key `discord_url` (`optional: true` in `console.yaml`). To enable after merge:

```bash
kubectl -n eurobase patch configmap eurobase-config --type merge \
  -p '{"data":{"discord_url":"https://discord.gg/XXXX"}}'
kubectl -n eurobase rollout restart deployment/console
```

For local dev, set `PUBLIC_DISCORD_URL` in `console/.env` (gitignored).

## Validation
- `npm run build` passes (both `login` and `(app)/_layout` compile).
- `console.yaml` YAML validated.

## Sovereignty
Consistent with the EU-only positioning: the link is opt-in and the disclosure makes the US-processor boundary explicit. No Eurobase/tenant data leaves the EU via this link — it only points users to the community server.


STGime:
## Code review done (high-effort, recall pass) — fixes in `4e59295`

**No correctness bugs.** The main risk I checked — `ssr = false` on `(app)` breaking `$env/dynamic/public` in the browser — was verified as a **non-issue**: adapter-node serializes public env into the CSR shell at server startup, so both links render in prod when `PUBLIC_DISCORD_URL` is set. (`rel="noopener noreferrer"`, `aria-hidden`, and `optional: true` on the configMapKeyRef all correct.)

**One real finding, fixed:** the compliance disclosure text had already **drifted** between the two files (login: *"send…to Discord, only what you choose to post"* vs sidebar: *"share…, only what you post"*) — two sources of truth for a legal notice. Fixes:

- `$lib/discord.ts` — single canonical `DISCORD_DISCLOSURE` string, used by both the login `<p>` and the sidebar `title`.
- `$lib/DiscordIcon.svelte` — shared icon component, replacing the duplicated inline SVG in both files.
- login link gets `aria-describedby` so screen readers announce the US-service notice.

Console build passes after the refactor.

PR #194, opened 2026-06-10
Fix #193: embed IANA tzdata in gateway and worker binaries

### LINKED ISSUE #193: Schedules API rejects valid IANA timezones (Europe/Berlin) — container likely missing tzdata
`eb.functions.schedules.createOrUpdate(name, { cron, timezone: 'Europe/Berlin' })` fails with:

```
invalid timezone "Europe/Berlin": unknown time zone Europe/Berlin
```

`schedules.ts` documents `timezone` as "Optional IANA timezone. Defaults to UTC." `Europe/Berlin` is about as canonical as IANA gets, so this looks like Go's `time.LoadLocation` failing inside a container without tzdata.

Fix is usually one of:
- `import _ "time/tzdata"` in the gateway binary (Go ≥1.15 embedded zoneinfo), or
- install tzdata in the gateway image.

Workaround: `timezone: 'UTC'`.

STGime:
## Problem

`eb.functions.schedules.createOrUpdate(name, { cron, timezone: 'Europe/Berlin' })` failed with `invalid timezone "Europe/Berlin": unknown time zone Europe/Berlin` (#193).

The schedules API validates timezones with `time.LoadLocation`, which reads `/usr/share/zoneinfo` at runtime. Both runtime images (`Dockerfile.gateway`, `Dockerfile.worker`) are bare `alpine:3.20` without the `tzdata` package, so every non-UTC IANA timezone was rejected.

## Fix

Blank-import `time/tzdata` in `cmd/gateway/main.go` and `cmd/worker/main.go` to embed the IANA zone database at build time (~450 KB per binary):

- **gateway** — validates `timezone` on schedule create/update (`internal/cron/service.go`)
- **worker** — resolves the timezone when computing next-run times in the cron executor (`internal/cron/executor.go`)

Chose the embed over installing the `tzdata` apk so the fix can't regress on a future base-image change and needs no Dockerfile edits.

## Verification

`go build` + `go vet` on both binaries pass. After deploy: `schedules.createOrUpdate` with `timezone: 'Europe/Berlin'` should be accepted and next-run times computed in local time.

Closes #193

### DIFF
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -12,6 +12,9 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	// Embed IANA tzdata so time.LoadLocation works in the alpine runtime
+	// image, which ships no /usr/share/zoneinfo (schedules API timezones).
+	_ "time/tzdata"
 
 	"github.com/eurobase/euroback/internal/auth"
 	"github.com/eurobase/euroback/internal/compliance"
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -9,6 +9,9 @@ import (
 	"syscall"
 
 	"time"
+	// Embed IANA tzdata so time.LoadLocation works in the alpine runtime
+	// image, which ships no /usr/share/zoneinfo (cron schedule timezones).
+	_ "time/tzdata"
 
 	"github.com/eurobase/euroback/internal/cron"
 	"github.com/eurobase/euroback/internal/db"


STGime (commented):
## Review — Approve

Correct, minimal fix. Verified all three `time.LoadLocation` call sites the embed actually serves:

- gateway validates the timezone on schedule create/update — `internal/cron/service.go:80` and `:220`
- worker resolves it for next-run computation — `internal/cron/executor.go:70`

So both blank imports are load-bearing, not dead weight. Embedding `time/tzdata` over an apk `tzdata` is the right call for the stated reason — it can't regress on a future base-image swap and needs no Dockerfile edit. ~450 KB/binary is negligible.

### Nits (non-blocking)
- `cmd/worker/main.go` has a pre-existing stray blank line between `"syscall"` and `"time"`; the new import groups fine, but worth a `gofmt -l ./cmd/...` to confirm the comment+import block is canonical (gofmt-clean, not just `vet`-clean).
- Consider a one-line regression guard so a future build-tag/CGO change can't silently drop tzdata again:
  ```go
  func TestEuropeBerlinResolvable(t *testing.T) {
      if _, err := time.LoadLocation("Europe/Berlin"); err != nil { t.Fatal(err) }
  }
  ```
  Cheap insurance for exactly the bug this closes.

No correctness/security/perf concerns. LGTM.

Issue #189, opened 2026-06-10
Edge functions advertise TypeScript/Deno DX but runner only executes plain JS via new Function()


STGime:
Found while building predwell on Eurobase.

## Problem

The docs and schema promise TypeScript: `docs/architecture-edge-functions.md` says "Built-in TypeScript (no build step) … same DX as Supabase Edge Functions", and `edge_functions.code` is described as "TypeScript/JavaScript source".

In reality `worker_bootstrap.js` loads user code with `new Function(wrapped)()` inside a `permissions: 'none'` worker, which means:

- **TypeScript syntax throws at load time** (no transpile step anywhere in the path)
- **`import` / `export default` are unsupported** — only `globalThis.handler = …`, `exports.default`, or `module.exports` shapes are detected (the bootstrap itself notes it "cannot detect `export default` syntax")
- No third-party libraries at all (no deno-dom, no SDK) — e.g. predwell's WG-Gesucht scraper has to parse HTML with regexes

Anyone pasting a Supabase-style `import { serve } from …` / `export default` TS function gets an opaque "Function must export a default handler" or a syntax error.

## Also: stale sandbox comment

`worker_bootstrap.js` still says *"net: false → user JS cannot fetch"*, but `server.ts` (closes #83) deliberately allows outbound net for external APIs. The comment cost me a near-rearchitecture before reading `server.ts`. Worth fixing the comment when touching this file.

## Suggested fix (either direction)

1. Transpile on deploy (esbuild/`Deno.emit` strip-types) and load code as a real module inside the worker so TS + ES module syntax work as documented; or
2. Update docs, console templates, and CLI deploy errors to state the actual contract: plain JS, `globalThis.handler = async (req, ctx) => …`, no imports — and validate/reject TS or module syntax at deploy time with a clear message instead of failing at first invocation.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

PR #195, opened 2026-06-10
Fix #188: set end-user/service RLS GUCs in function-runner SQL

### LINKED ISSUE #188: Function runner never sets end-user/service GUCs — RLS blocks edge-function DB access

## Problem

`functions-runner/server.ts` runs each `ctx.db.sql` call with only:

```ts
const setRoleSQL = "SET LOCAL ROLE " + quoteIdent(funcRole);
const setPathSQL = "SET LOCAL search_path TO " + quoteIdent(schemaName);
```

It never sets `app.end_user_id` or `app.end_user_role`. Consequences inside any edge function:

- `public.is_service_role()` is **false** (defined in 000038 as `current_setting('app.end_user_role') = 'service'`)
- `auth_uid()` is **NULL**, even when the function is deployed with `verify_jwt=true` and the gateway forwards `X-User-ID`

So edge functions satisfy *no branch* of the RLS policies the platform's own DDL presets generate (`public.is_service_role() OR <owner> = auth_uid()`). A function can't write to — or even read — any preset-protected table, and `verify_jwt` functions can't act on behalf of the authenticated user at the SQL level.

## Repro

1. Create a table via the DDL endpoint with the default `owner_access` preset.
2. Deploy a function that does `ctx.db.sql("INSERT INTO t (user_id, ...) VALUES ($1, ...)", [ctx.user.id])`.
3. Invoke with a valid end-user JWT → `row-level security policy denied this operation`.

## Workaround used in predwell

Every "service" RLS branch additionally accepts the runner role by name:

```sql
(public.is_service_role() OR current_user = 'tenant_<id>_func')
```

…but tenants shouldn't need to know the runner's role naming convention.

## Suggested fix

In `runDBSql`, alongside `SET LOCAL ROLE`, set the same GUCs the gateway sets:

- `SELECT set_config('app.end_user_id', $userId, true)` when `X-User-ID` is present
- `set_config('app.end_user_role', 'authenticated' | 'service', true)` (service when no end-user JWT, mirroring gateway service paths)

That makes `auth_uid()` / `is_service_role()` behave identically in functions and in gateway REST, and the DDL presets become usable from functions with zero tenant workarounds.

STGime:
## Problem

The runner's per-invocation transaction set only `SET LOCAL ROLE <schema>_func` + `search_path`, never `app.end_user_id` / `app.end_user_role`. Inside any edge function:

- `public.is_service_role()` was **false** (it checks `app.end_user_role = 'service'`, migration 000038)
- `auth_uid()` was **NULL**, even for `verify_jwt=true` functions where the gateway forwards `X-User-ID`

So `ctx.db.sql` satisfied **no branch** of the RLS policies the platform's DDL presets generate (`public.is_service_role() OR <owner> = auth_uid()`) — functions couldn't read or write preset-protected tables at all, and tenants had to hand-write policy branches naming the runner's role convention (#188).

## Fix

`runDBSql` now mirrors the gateway's RLS context (`internal/query/engine.go` → `applyRLSContext`), via a new testable `rlsContextStatements` helper in `role.ts`:

- **`X-User-ID` present** → `app.end_user_id = <user>`, `app.end_user_role = 'authenticated'`. The header is trustworthy: the gateway only sets it from verified end-user claims, and the HMAC signature covers it.
- **No end-user** → `app.end_user_role = 'service'`, mirroring gateway secret-key traffic — function code is developer-authored server-side code.

The `anon` branch is deliberately not mirrored: anonymous end-user semantics make sense for client-issued queries, not tenant-deployed code.

The user id always travels as a bind parameter, never interpolated into SQL text.

## Tests

New Deno tests in `role_test.ts` cover the authenticated branch, the service branch, and an injection-shaped user id (same local-runnable pattern as the existing role tests; CI doesn't run Deno tests). Syntax verified with esbuild.

After deploy, the repro from the issue (DDL-preset table + function inserting with `ctx.user.id` under a valid end-user JWT) should succeed instead of `row-level security policy denied this operation`, and the `current_user = 'tenant_<id>_func'` workaround branches can be dropped.

Closes #188

### DIFF
--- a/functions-runner/role.ts
+++ b/functions-runner/role.ts
@@ -15,9 +15,11 @@
 // SQL via double-quote wrapping. Letters, digits, underscores; leads
 // with letter/underscore; max 63 bytes (Postgres's NAMEDATALEN-1
 // default). Schemas the gateway forwards always conform to this shape;
-// the runner validates again so a forged X-Schema-Name header (until
-// HMAC verification ships in PR 3c) cannot smuggle quote characters or
-// SQL fragments.
+// gateway→runner traffic is HMAC-signed and verified (see
+// authenticateRequest in server.ts), and the runner validates the shape
+// again as defence-in-depth so a forged X-Schema-Name header cannot
+// smuggle quote characters or SQL fragments even if that signing is ever
+// weakened by misconfiguration.
 export const validIdentRe = /^[a-zA-Z_][a-zA-Z0-9_]{0,62}$/;
 
 // tenantFuncRole returns the per-tenant Postgres role name. Created at
@@ -51,6 +53,15 @@ export function quoteIdent(s: string): string {
 // the service branch — same as gateway secret-key traffic. The 'anon'
 // branch is deliberately not mirrored: anonymous *end-user* access only
 // makes sense for client-issued queries, not for tenant-deployed code.
+//
+// SECURITY: X-User-ID now drives the RLS *identity* (app.end_user_id +
+// the authenticated role), not just ctx.user, so a forged value would
+// impersonate that user at the row-security layer for every table. This
+// is only safe because gateway→runner traffic is HMAC-signed (X-User-ID
+// is in the canonical message) AND the runner runs in strict mode
+// (FUNCTIONS_RUNNER_HMAC_REQUIRE_SIGNED=true in deploy/k8s/functions.yaml).
+// Do not relax HMAC verification to soft mode without accounting for the
+// fact that it now governs database row security, not only ctx.user.
 export function rlsContextStatements(
   userId: string,
 ): Array<{ sql: string; params: string[] }> {
--- a/functions-runner/role.ts
+++ b/functions-runner/role.ts
@@ -38,3 +38,38 @@ export function tenantFuncRole(schemaName: string): string {
 export function quoteIdent(s: string): string {
   return '"' + s.replaceAll('"', '""') + '"';
 }
+
+// rlsContextStatements returns the per-transaction set_config statements
+// that mirror the gateway's RLS context (internal/query/engine.go,
+// applyRLSContext) so auth_uid() / is_service_role() behave identically
+// in edge functions and gateway REST (closes #188).
+//
+// With an end-user (gateway sets X-User-ID from verified claims, covered
+// by the HMAC signature) the function acts as that user: policies see
+// app.end_user_role='authenticated' and auth_uid() = the user. Without
+// one, function code is developer-authored server-side code, so it gets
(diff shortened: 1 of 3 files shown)


STGime (commented):
## Review — Approve, with one note to surface

Correct and well-scoped. Traced the data path: `userId` is read in `executeFunction` (server.ts:198), `runDBSql` is a closure over it, and the new statements run after `SET LOCAL ROLE`/`search_path` inside each per-call `db.begin`. Since `set_config(..., true)` is transaction-local and every `runDBSql` is its own transaction, the GUCs re-apply per statement and can't leak across pooled-connection reuse — exactly right.

Other things I checked and liked:
- User id travels as a bind param, never interpolated; the injection-shaped test locks that.
- Branch design matches `internal/query/engine.go` `applyRLSContext`; deliberately not mirroring the `anon` branch is the correct call for server-side code.
- `X-User-ID` **is** in the HMAC canonical message (`hmac.ts` `canonicalMessage`, `user=` field), `authenticateRequest` runs before `executeFunction` (server.ts:583), and `deploy/k8s/functions.yaml` sets `FUNCTIONS_RUNNER_HMAC_REQUIRE_SIGNED: "true"`. So the "header is trustworthy" claim genuinely holds in prod.

### The note (not blocking)
This PR **widens the blast radius of a forged `X-User-ID`**. Before, a spoofed header only shaped `ctx.user` (app-level, function may or may not trust it). Now it drives `app.end_user_id` + the authenticated role at the **RLS layer** for every preset-protected table. That's safe *because* the header is HMAC-signed and the runner is in strict mode — but the safety of every tenant's row security now depends on that staying true. Two small asks:

1. Make the dependency explicit — a line in the PR body or a comment in `rlsContextStatements` noting that running the runner in soft mode (or with `HMAC_SECRET` unset) would let an intra-cluster forged `X-User-ID` assume DB identity. Future operators shouldn't be able to relax HMAC without realizing it now governs RLS, not just `ctx.user`.
2. Stale comment: `server.ts:209` still says *"cf. PR 3c which adds HMAC verification of these headers"* — verification has landed, so that now misreads as "not yet protected." Worth deleting/updating (same stale-comment class as #189). The `role.ts:19` "until HMAC verification ships in PR 3c" is similarly stale.

Test coverage on the helper is good; the only untested seam is the integration (that the GUCs actually flip `auth_uid()`/`is_service_role()` under the func role) — acknowledged as not CI-runnable, fine.

Confirmed against predwell: after deploying the equivalent fix the `current_user = 'tenant_<id>_func'` workaround branches can indeed be dropped. LGTM.

PR #196, opened 2026-06-10
Fix #192: CLI status/deploy must show and validate the active project

### LINKED ISSUE #192: CLI: status shows empty project after login restores one; deploys land in stale active project unannounced
Found while deploying predwell functions.

## What happened

Fresh CLI login, **no active project selected** (`eurobase status` printed an empty `Project:  ()`). Running

```
eurobase edge-functions deploy free-report -f functions/free-report.js --no-verify-jwt
```

printed `✓ Function "free-report" created and deployed` — but the function did not appear in the intended project. It had been created in a *different* project (`newtek`, 93df6379), presumably whatever the token/context fell back to. Invoking it on the intended project returned `{"error":"function not found"}` with no hint why.

After `eurobase switch predwell`, the same deploy command worked correctly.

## Expected

- With no active project set, mutating commands (deploy/delete/cron/vault) should **fail fast**: "no active project — run `eurobase switch <project>`", not pick a fallback and report success.
- Deploy output should always name the target: `✓ Function "free-report" deployed to predwell (0515e4e2…)` — that alone would have made the misfire obvious.

## Related polish

`eurobase status` prints `Project:  ()` (empty name/plan) even immediately after a successful `eurobase switch predwell` — so the usual way to double-check the active context doesn't help.

STGime:
## Root cause

Not a missing-project fallback — `RequireProject` was already enforced on every mutating command. Two compounding bugs made the misfire invisible (#192):

1. **`login` keeps a stale `active_project`.** It only auto-selects when the account has exactly one project; otherwise it silently leaves whatever was in `~/.eurobase/config.json` — even a project from another account or a deleted one. That's how the deploy landed in `newtek`.
2. **`eurobase status` was unusable for checking context.** It parsed a flat `{project_name, db_usage_mb, ...}` shape, but the endpoint (`internal/plans/handler.go` → `HandleGetUsage`) returns `{usage: {...}, limits: {...}}` — nothing matched, so it always printed `Project:  ()` with zeroed bars, for everyone.

## Fixes

- **login** reconciles the stored selection against the account's project list (new `ReconcileActiveProject`): single-project accounts auto-select as before; a selection not in the list is **cleared**, so subsequent mutating commands fail fast with ``no active project — run `eurobase switch <slug>` first``; a valid selection gets its slug healed. Login prints the active project, or `No active project — run `eurobase switch <slug>` to select one.`
- **status** parses the real `{usage, limits}` shape (plan from limits) and shows the active project from the local config via the new `ProjectLabel` helper — slug + ID.
- **edge-functions deploy/delete** output names the target: `✓ Function "free-report" deployed to predwell (0515e4e2…)`.

## Tests

First unit tests for `internal/cli`: reconciliation (auto-select, stale-clear, slug-heal, multi-project no-select, empty list) and `ProjectLabel`. `go build` / `go vet` / `go test ./internal/cli/` pass.

Closes #192

### DIFF
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -83,17 +83,16 @@ func LoginCmd() *cobra.Command {
 			cfg.Token = result.AccessToken
 			cfg.Email = email
 
-			// Try to fetch projects and auto-select if only one
+			// Fetch projects: auto-select a single-project account, and
+			// validate any previously stored active project against this
+			// account's list so a stale selection (other account, deleted
+			// project) can't silently receive deploys (issue #192).
 			client.Token = result.AccessToken
 			projectsData, err := client.Get("/v1/tenants")
 			if err == nil {
-				var projects []struct {
-					ID   string `json:"id"`
-					Slug string `json:"slug"`
-				}
-				if json.Unmarshal(projectsData, &projects) == nil && len(projects) == 1 {
-					cfg.ActiveProject = projects[0].ID
-					cfg.ProjectSlug = projects[0].Slug
+				var projects []ProjectRef
+				if json.Unmarshal(projectsData, &projects) == nil {
+					ReconcileActiveProject(cfg, projects)
 				}
 			}
 
@@ -103,7 +102,9 @@ func LoginCmd() *cobra.Command {
 
 			PrintSuccess(fmt.Sprintf("Logged in as %s", email))
 			if cfg.ActiveProject != "" {
-				PrintSuccess(fmt.Sprintf("Active project: %s", cfg.ProjectSlug))
+				PrintSuccess(fmt.Sprintf("Active project: %s", ProjectLabel(cfg)))
+			} else {
+				fmt.Println("No active project — run `eurobase switch <slug>` to select one.")
 			}
 			return nil
 		},
--- a/internal/cli/config.go
+++ b/internal/cli/config.go
@@ -188,3 +188,45 @@ func RequireProject(cfg *Config) error {
 	}
 	return nil
 }
+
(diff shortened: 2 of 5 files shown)


STGime (commented):
## Review — Approve

Root-cause diagnosis is right and both fixes are clean. Verified the status.go shape against the server: `internal/plans/handler.go` returns `{usage, limits}`, and the new struct's tags match exactly — `database_size_mb`/`storage_size_mb`/`mau_count` (`usage.go`) and `plan`/`db_size_mb`/`storage_mb`/`mau_limit` (`limits.go`). So the `Project:  ()` + zeroed-bars bug is genuinely fixed, and it was hitting *every* user, not just multi-project accounts — good catch.

`ReconcileActiveProject` logic is sound across all branches; I walked the empty-list case specifically: `len != 1`, `ActiveProject != ""` so no early return, empty loop falls through to clear — matches `TestReconcileActiveProject_EmptyListClearsSelection`. The six tests are all meaningful (nice to see the first `internal/cli` unit coverage), and `ProjectLabel` on deploy/delete output is exactly what would have made my original misfire visible.

### Nits (non-blocking)
- `status` now sources the project label from local config (`ProjectLabel(cfg)`) but the usage numbers from the server. `status` itself doesn't call `ReconcileActiveProject`, so a stale config slug could in principle show a label that disagrees with the usage shown. Low risk now that `login` heals the slug, but if you want belt-and-braces, have `status` fall back to the server's project identity (or note the label is config-sourced).
- Optional follow-up: print the active project (or the "no active project — run `switch`" hint) on `switch` confirmation too, so every context-changing command is symmetric.

LGTM.

PR #197, opened 2026-06-10
Fix #189: transpile TypeScript/ESM edge functions on deploy (esbuild)

### LINKED ISSUE #189: Edge functions advertise TypeScript/Deno DX but runner only executes plain JS via new Function()

STGime:
## Problem

`docs/architecture-edge-functions.md` promised "Built-in TypeScript (no build step) … same DX as Supabase Edge Functions", and `edge_functions.code` is documented as TypeScript/JavaScript source. In reality the runner executes user code with `new Function()` inside a permissions-locked worker: TypeScript syntax threw at load time, `export default` was undetectable, and anyone pasting a Supabase-style function got an opaque "Function must export a default handler" on first invocation (#189).

## Approach: compile at deploy time, not runtime

One esbuild transform (`LoaderTS` + `FormatCommonJS`) kills both bugs without touching the sandbox: types are stripped and `export default` is rewritten into the `module.exports` shape the worker bootstrap detects. The worker's permission set is unchanged.

### Gateway

- **`internal/functions/transpile.go`** — esbuild Go API, in-process, ~ms per deploy. Compile errors fail the deploy with line/column diagnostics instead of an opaque first-invocation failure. Note: this is type-**stripping**, not type-checking — a `tsc` error can still deploy.
- **`Service.Create` / `Update` / `Rollback`** compile on every code change. Metadata-only updates (status toggle, env vars) never recompile, so a legacy function that no longer compiles can still be disabled. Versions store source only; rollback recompiles it fresh.
- **Migration `000059`** adds `edge_functions.compiled_code`: `code` keeps the original source (console editing, `ef get`), `compiled_code` is what the runner executes. No new grants — table-level grants cover new columns automatically.

### Runner

- `loadFunction` selects `COALESCE(compiled_code, code)` — NULL means a pre-transpile function, which is plain JS by contract, so existing functions keep working untouched.
- **`worker_bootstrap.js`**: esbuild *replaces* `module.exports` with a fresh object, so the existing `exports.default` check never sees its output — added a `module.exports.default` detection case. Added a `require()` stub that rejects third-party imports at load time with a clear "inline the dependency" message (ESM imports compile to `require()` calls; the sandbox has no module loader). Fixed the stale "net: false" comment that contradicted `server.ts` (called out in the issue).

### Docs / console

Architecture doc now states the real contract (deploy-time type-strip, no URL/npm imports yet). The console new-function template and docs page show `export default` + TypeScript instead of warning against them, with the third-party-import limitation called out.

## Known limitation

Third-party imports (`import { serve } from "https://deno.land/..."`, npm) are still unsupported — that needs module-graph bundling, out of scope here. They now fail with a clear load-time error instead of a `ReferenceError`.

## Verification

- Go unit tests pin the transpile contract: TS strip, `export default` shape, plain-JS passthrough, line-info errors, import→require. All pass.
- Real esbuild output run through the exact bootstrap wrapper in Node: `export default` detected, require stub throws the right message, legacy `globalThis.handler` functions still load.
- Deno sandbox tests added for esbuild-shape detection and the require stub (local-runnable; CI doesn't run Deno tests). Syntax verified with esbuild.
- Console `npm run check` errors are all pre-existing baseline, none at the edited lines.

Closes #189

## Deploy ordering (review note)

This is an expand-style change with one ordering dependency: **migration `000059` must apply before the new runner image rolls out** — the runner's `SELECT COALESCE(compiled_code, code)` errors if the column doesn't exist yet. The reverse is safe: with the migration applied and the old runner still live, the old runner keeps selecting `code` and ignores the new column. CI applies migrations via the `migrate` Job before rolling Deployments, so the normal pipeline ordering already satisfies this — it only matters for manual/partial rollouts or rollbacks of the migration.


STGime (commented):
## Review — Approve (this is the fix for the issue I filed; thanks)

Compile-at-deploy is the right call over runtime transpile — the sandbox is untouched, cold-path cost is paid once per deploy, and `code`/`compiled_code` split keeps console editing + `ef get` showing real source. I checked this out and verified the load-bearing claims rather than trusting the diff:

**Verified locally (Go 1.26):**
- `go build` / `go test` / `go vet ./internal/functions/` all pass.
- **The two halves actually mesh** — I ran real `esbuild --format=cjs` output through the exact `worker_bootstrap.js` wrapper + `HANDLER_DETECT` suffix in Node:
  - `export default async (req,ctx)=>…` → handler detected ✅
  - `export default function handler(){}` → detected ✅
  - `import {serve} from "https://…"` → throws the require-stub message at load ✅
  - plain `globalThis.handler = …` (legacy) → still detected ✅
  So the new `module.exports.default` detection case correctly catches esbuild's replace-module.exports shape, and the require stub is the right backstop for the import→require compile path.
- Migration `000059` is correctly the latest (follows `000058`), down-migration is idempotent.

**One real nit (CI will catch it):** `esbuild` is in the `// indirect` block of `go.mod`, but it's a direct import (`internal/functions/transpile.go`). `go mod tidy` moves it into the direct `require` block and drops `// indirect` — so a `go mod tidy && git diff --exit-code` lint step will fail. Run tidy and commit the result.

**One operational note (not a code bug):** the runner's `SELECT COALESCE(compiled_code, code)` requires the `compiled_code` column to exist. That makes this an expand-style change with a deploy-ordering dependency: **migration 000059 must apply before the new runner image rolls out**, or `loadFunction` errors with "column compiled_code does not exist". The reverse (migration applied, old runner still live) is safe — old runner selects `code` and ignores the new column. Worth a line in the deploy runbook / PR body so the rollout order is explicit.

**Minor observations (optional):**
- `Loader: LoaderTS` won't handle JSX/TSX. Almost certainly fine for edge functions, but if someone pastes JSX they'll get a compile error rather than support — acceptable, maybe one line in docs.
- Type-strip-not-check is clearly documented in `transpile.go`, the migration comment, and the docs — good. A `tsc` error still deploying is the right tradeoff for a server-side deploy path.
- Nice touch recompiling on rollback so a legacy version that no longer compiles fails loudly with the diagnostic instead of shipping stale output.

Test coverage is solid (transpile contract + sandbox shape + require stub), and the stale `net: false` bootstrap comment I flagged in #189 is fixed here too. LGTM once `go mod tidy` is run.

PR #202, opened 2026-06-11
Fix #188 (follow-up): grant func roles access to public RLS helpers

### LINKED ISSUE #188: Function runner never sets end-user/service GUCs — RLS blocks edge-function DB access

STGime:
## Problem

PR #195 fixed the GUC half of #188, but live testing on predwell surfaced a second gap: migration 000047 granted `USAGE ON SCHEMA public` and EXECUTE on the RLS helper functions only to `eurobase_function_runner` — and the per-tenant `<schema>_func` role does **not** inherit those (membership goes runner → func role, not the other way). After `SET LOCAL ROLE <schema>_func`, any RLS policy that merely *references* `public.is_service_role()` or `public.current_end_user_id()` — i.e. every policy the platform's DDL presets generate — fails at expression init with `permission denied for schema public`.

## Fix — migration 000060

1. **Backfill**: grants `USAGE ON SCHEMA public` + `EXECUTE` on `is_service_role()`, `current_end_user_id()`, and `is_internal_auth_path()` to every existing tenant's func role.
2. **provision_tenant**: re-created with the 000057 body + the same grants, so new tenants get them at provisioning time (the established wholesale-re-create pattern from 000050/52/55/56/57).

`is_internal_auth_path()` is included so policies on the sensitive auth tables (000055) evaluate to a clean `false`-deny for the func role instead of erroring. None of the grants leak data: the helpers only read transaction-local GUCs, and `USAGE` on `public` alone confers no table access (the func roles have no table grants there).

## Verification

- Down migration restores the 000057 `provision_tenant` body — **diff-verified byte-identical**.
- Up migration's `provision_tenant` is 000057's body plus exactly the four grant lines — diff-verified.
- No live Postgres available locally; the CI migrate Job gates the rollout (fails before any Deployment rolls).

After deploy, predwell's `current_user = 'tenant_<id>_func'` policy workarounds can be dropped and the DDL-preset policies work from edge functions.

Closes #188

### DIFF
--- a/migrations/000060_func_role_public_grants.down.sql
+++ b/migrations/000060_func_role_public_grants.down.sql
@@ -1,0 +1,253 @@
+-- Reverts the issue #188 follow-up grants and restores provision_tenant
+-- to the 000057 body. Note: revoking re-breaks RLS policy evaluation for
+-- edge functions (the bug this migration fixes).
+
+-- ── 1. Revoke from existing tenants' func roles ────────────────────────
+DO $$
+DECLARE
+    v_schema TEXT;
+    v_role   TEXT;
+BEGIN
+    FOR v_schema IN
+        SELECT schema_name FROM public.projects WHERE schema_name IS NOT NULL
+    LOOP
+        v_role := v_schema || '_func';
+        IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = v_role) THEN
+            EXECUTE format('REVOKE EXECUTE ON FUNCTION public.is_internal_auth_path() FROM %I', v_role);
+            EXECUTE format('REVOKE EXECUTE ON FUNCTION public.current_end_user_id() FROM %I', v_role);
+            EXECUTE format('REVOKE EXECUTE ON FUNCTION public.is_service_role() FROM %I', v_role);
+            EXECUTE format('REVOKE USAGE ON SCHEMA public FROM %I', v_role);
+        END IF;
+    END LOOP;
+END$$;
+
+-- ── 2. Restore provision_tenant to the 000057 body ─────────────────────
+CREATE OR REPLACE FUNCTION public.provision_tenant(
+    p_project_id   UUID,
+    p_display_name TEXT,
+    p_plan         TEXT DEFAULT 'free'
+)
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+AS $fn$
+DECLARE
+    v_schema_name TEXT;
+    v_func_role   TEXT;
+BEGIN
+    v_schema_name := 'tenant_' || replace(p_project_id::text, '-', '_');
+    v_func_role   := v_schema_name || '_func';
+
+    EXECUTE format('CREATE SCHEMA %I', v_schema_name);
+    EXECUTE format('SET search_path TO %I', v_schema_name);
+
+    EXECUTE format(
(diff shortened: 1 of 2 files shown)


STGime (commented):
## Review — Approve

This is the exact gap I hit on predwell after #195 — functions got `permission denied for schema public` because the `<schema>_func` role doesn't inherit the runner's grants. The fix is right and minimal.

**Verified mechanically (not just by eye):**
- Diffed the up migration's `provision_tenant` body against `000057_vault_key_version.up.sql` — the **only** difference is the 4 inserted lines (`GRANT USAGE ON SCHEMA public` + EXECUTE on `is_service_role` / `current_end_user_id` / `is_internal_auth_path`), placed right before `GRANT <func_role> TO eurobase_function_runner`. No drift in the wholesale re-create.
- Diffed the down migration's `provision_tenant` against 000057 → **identical**. The "byte-identical restore" claim holds.
- Backfill loop is idempotent (GRANT is idempotent; `IF EXISTS` guard on the role).

Good call including `is_internal_auth_path()` so the sensitive auth-table policies (000055) evaluate to a clean `false`-deny under the func role rather than erroring. The grants genuinely don't leak data — USAGE on `public` confers no table access, and the helpers only read transaction-local GUCs.

Once this lands I can drop predwell's `current_user = 'tenant_<id>_func'` policy workarounds (migrations 000008–000011) and use the stock DDL presets. LGTM.

PR #203, opened 2026-06-11
Fix #200: key the runner code cache by function version

### LINKED ISSUE #200: Runner serves stale function code after redeploy — deploy doesn't invalidate the in-memory LRU (up to 5-min TTL)
Found while iterating on predwell edge functions.

## Symptom

Redeployed a function (`eurobase edge-functions deploy notify-matches …`), the CLI confirmed success, and `edge-functions get` showed the new source at **version 2**. But invoking it kept executing the **old (v1) code** — it threw an error that only the v1 SQL could produce, identical across repeated calls. ~5 minutes later, with no further action, the same invocation started running v2 and succeeded.

A brand-new function deployed at the same time with the *identical* v2 SQL ran correctly on first call — because it had never been cached.

## Root cause

`functions-runner` caches loaded function code in an in-memory LRU with a 5-minute TTL (per `docs/architecture-edge-functions.md` §3.3, and `loadFunction` in `functions-runner/server.ts`). The cache appears keyed by function id/name and refreshed only on TTL expiry — a redeploy bumps `edge_functions.version` (and `compiled_code`, post-#197) in the DB but does **not** invalidate the runner's cached entry. So for up to 5 minutes after a deploy, invocations run stale code while the dashboard/`get` show the new version. Very confusing during iteration — you change code, redeploy, and the bug "doesn't go away."

## Suggested fixes (any one)

1. **Cache-key by version.** Include `version` (or `updated_at`) in the cache key, or store it alongside the entry and re-fetch when the DB row's version is newer. `loadFunction` already hits the DB to read code on a miss; a cheap `SELECT version` check (or returning version with the code and comparing) closes the window.
2. **Invalidate on deploy.** Have the gateway's create/update/rollback path signal the runner to evict that function (internal HTTP call, or a pub/sub/`LISTEN` channel) so redeploys take effect immediately.
3. **At minimum, document it** in the CLI deploy output ("changes take effect within ~5 min as the runner cache refreshes") so the staleness isn't mistaken for a failed deploy or a code bug.

Option 1 is the least surprising — a deploy should be immediately effective, which is the normal expectation for serverless platforms.

STGime:
## Problem

The runner caches function code by id with a 5-minute TTL and nothing invalidates it on deploy — so for up to 5 minutes after a redeploy, invocations run stale code while the CLI/dashboard show the new version (#200). Post-#197 this also means stale *compiled* output.

## Fix (issue's option 1)

Both invocation paths already look up the function row, so they now pass the version they fetched:

- gateway `HandleInvoke` sends `X-Function-Version` from `fn.Version`
- the worker's cron invoker adds `version` to its existing id/status lookup and sends the same header

The runner keys its cache on `<id>@v<version>` (new `cache_key.ts`), so a version bump is a guaranteed cache miss — redeploys take effect immediately. Old-version entries age out via the existing LRU/TTL.

## Rollout / security notes

- The header is deliberately **not** HMAC-covered: a forged version can only cause a cache miss — code is always loaded from the DB by id, so there's no poisoning path. Non-numeric values are ignored.
- Missing version (older gateway during a rolling deploy) falls back to the id-only key — the previous TTL-bounded behaviour — so gateway/runner rollout order doesn't matter.

## Tests

`cache_key_test.ts` covers version-busts-cache, id-only fallback, and garbage-version handling (runs in CI's test-functions-runner job). `go build`/`vet`/`test` pass for the gateway and cron packages.

Closes #200

### DIFF
--- a/functions-runner/cache_key.ts
+++ b/functions-runner/cache_key.ts
@@ -1,0 +1,19 @@
+// Cache-key derivation for the runner's in-memory code cache.
+//
+// Closes #200: the cache was keyed by function id alone with a 5-minute
+// TTL, so a redeploy kept serving stale code until the TTL expired. The
+// gateway (and the worker's cron invoker) now send X-Function-Version —
+// already looked up for the invoke — and the key includes it, so a
+// version bump is a guaranteed cache miss and redeploys take effect
+// immediately. Old-version entries age out via the existing LRU/TTL.
+//
+// The header is not HMAC-covered: a forged version can only cause a
+// cache miss — code is always loaded from the DB by id. An absent or
+// garbage version (older gateway during a rolling deploy) falls back to
+// the id-only key, i.e. the pre-#200 TTL-bounded behaviour.
+export function functionCacheKey(functionId: string, version: string | null): string {
+  if (version && /^[0-9]+$/.test(version)) {
+    return `${functionId}@v${version}`;
+  }
+  return functionId;
+}
--- a/functions-runner/cache_key_test.ts
+++ b/functions-runner/cache_key_test.ts
@@ -1,0 +1,25 @@
+// Tests for the code-cache key derivation (issue #200): a version bump
+// must produce a different key (immediate redeploy effect), while a
+// missing/garbage version falls back to the id-only key.
+
+import { assertEquals, assertNotEquals } from "https://deno.land/std@0.224.0/assert/mod.ts";
+import { functionCacheKey } from "./cache_key.ts";
+
+const ID = "0a1b2c3d-0000-4000-8000-000000000001";
+
+Deno.test("version is part of the cache key — redeploy busts the cache (issue #200)", () => {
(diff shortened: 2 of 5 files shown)


STGime (commented):
## Review — Approve

Clean fix for the stale-cache surprise I filed. Verified both invocation paths send the version:
- gateway `internal/functions/invoke.go` → `X-Function-Version` from `fn.Version`
- cron `internal/cron/executor.go` → adds `version` to the existing id/status lookup and sets the same header

`cache_key.ts` keys on `<id>@v<version>` with a numeric-only guard and id-only fallback, so a redeploy is a guaranteed miss while an absent/garbage version (older gateway mid-rollout) degrades to the previous TTL-bounded behavior. The not-HMAC-covered reasoning is sound — a forged version only causes a cache miss; code is always loaded from the DB by id. `cache_key_test.ts` covers bust / fallback / garbage. Rollout order is genuinely irrelevant. LGTM.

PR #204, opened 2026-06-11
Fix #198: CORS-allow localhost:3000 by default + log rejected origins

### LINKED ISSUE #198: eurobase init scaffolds a localhost:3000 app but doesn't set cors_origins — every fresh browser app hits a CORS wall
Found while building predwell (a Vite/React app on Eurobase).

## Symptom

A freshly-scaffolded browser frontend calling `@eurobase/sdk` from `http://localhost:3000` gets an opaque failure on **every** call (auth, db, functions). In the app it surfaces as a generic "something went wrong"; in devtools it's a CORS error — the gateway sends no `Access-Control-Allow-Origin` because the project has no `cors_origins` configured. `curl` works fine (no CORS enforcement), so it's invisible in backend testing and only shows up the moment you open the real app in a browser.

## Root cause

`DefaultAuthConfig()` (`internal/tenant/auth_config.go`) sets `RedirectURLs: ["http://localhost:3000"]` but leaves `CORSOrigins` empty. And `eurobase init` scaffolds a project whose generated app/CLAUDE.md target `localhost:3000`. So the default project is set up to be called from `localhost:3000` for OAuth redirects but **not** for browser API calls — the two lists disagree for the exact origin the scaffold itself uses.

## Repro

1. `eurobase init` / create a project, build any browser app against the SDK on `localhost:3000`.
2. `eb.functions.invoke(...)` or `eb.db.from(...).select()` from the browser.
3. Browser blocks the response (no `access-control-allow-origin`); SDK reports a network error.
4. The same call via `curl` succeeds — masking the problem in tests.

## Suggested fixes (any one helps; first is most impactful)

1. **Default `cors_origins` to match `redirect_urls`** for new projects (at least include `http://localhost:3000`), or have `eurobase init` set it. If a project ships a `localhost:3000` redirect URL by default, the same origin should be CORS-allowed by default — they're set up for the same app.
2. **Make it discoverable from the CLI.** There's no `eurobase projects update` / config command — the only way to set `cors_origins` is `PATCH /v1/tenants/{id}` (admin token) or the console UI. A `eurobase projects set-cors <origin>` (or a general config-set) would make the fix obvious. Bonus: surface configured CORS origins in `eurobase status`.
3. **Better error signal.** When the gateway rejects/served a request whose `Origin` isn't allowlisted, it could log a one-line warning ("origin http://localhost:3000 not in cors_origins for project X") so the cause is greppable instead of a silent missing header.

## Workaround

Add the origin to `auth_config.cors_origins` via the console UI (Project → Auth settings) or:

```
PATCH /v1/tenants/{id}  { "auth_config": { …existing fields…, "cors_origins": ["http://localhost:3000"] } }
```

Note the PATCH **replaces** the whole `auth_config`, so all existing fields must be resent — easy to clobber `providers`/`session_duration` if you only send `cors_origins`. A partial/merge update (or a dedicated cors endpoint) would be safer.

STGime:
## Problem

`DefaultAuthConfig()` sets `RedirectURLs: ["http://localhost:3000"]` but left `CORSOrigins` empty — the exact origin the scaffold targets was trusted for OAuth redirects yet blocked for every browser API call (#198). A fresh Vite/React app fails on first fetch with no server-side signal: `curl` works (no CORS enforcement), so backend tests never see it.

## Fix

- **Defaults agree**: `DefaultAuthConfig` now sets `CORSOrigins` to match `RedirectURLs`. This applies to new projects and projects whose `auth_config` was never edited (`ParseAuthConfig` falls back to defaults for empty/`{}`/`null`). Projects with a stored config keep their explicit list — **no silent posture change on upgrade**. Keeping a localhost entry on production projects is not an exposure: remote sites cannot forge a localhost `Origin` header, and the SDK authenticates with bearer tokens, not ambient cookies.
- **Greppable signal** (issue fix 3): the CORS middleware logs one `slog.Warn` per rejected origin with `origin`, `project_id`, and `path` — previously the only symptom was a missing response header, invisible outside browser devtools.

## Tests

- every default redirect URL is CORS-allowed by the default config
- fresh project (empty auth_config) → preflight from `localhost:3000` succeeds
- stored config without `cors_origins` is **not** retroactively backfilled

## Follow-ups from the issue (not in this PR)

- `eurobase projects set-cors <origin>` CLI command (no CLI path to set CORS today)
- `PATCH /v1/tenants/{id}` replaces the whole `auth_config` — easy to clobber providers/session settings when only changing one field; deserves a merge-patch or dedicated endpoint

Closes #198

### DIFF
--- a/internal/gateway/cors.go
+++ b/internal/gateway/cors.go
@@ -1,6 +1,7 @@
 package gateway
 
 import (
+	"log/slog"
 	"net/http"
 	"strings"
 
@@ -49,6 +50,16 @@ func NewCORSMiddleware(allowedOrigins []string) func(http.Handler) http.Handler 
 				// Not allowed by global or per-project. For preflight
 				// respond 204 with no CORS headers so the browser
 				// rejects the real request.
+				//
+				// One greppable line per rejection (issue #198):
+				// without it the only symptom is a missing response
+				// header, invisible outside browser devtools.
+				projectID := ""
+				if pc, ok := auth.ProjectFromContext(r.Context()); ok && pc != nil {
+					projectID = pc.ProjectID
+				}
+				slog.Warn("CORS: origin not allowlisted, browser will block the response",
+					"origin", origin, "project_id", projectID, "path", r.URL.Path)
 				if r.Method == http.MethodOptions {
 					w.WriteHeader(http.StatusNoContent)
 					return
--- a/internal/gateway/cors_perproject_test.go
+++ b/internal/gateway/cors_perproject_test.go
@@ -140,3 +140,42 @@ func contains(haystack, needle string) bool {
 	}
 	return false
 }
+
+func TestCORS_FreshProjectWithEmptyAuthConfigAllowsLocalhost3000(t *testing.T) {
+	// Issue #198: a freshly-created project (empty/default auth_config)
+	// must be able to serve a localhost:3000 browser app out of the box —
+	// the scaffold and the default redirect_urls both target that origin.
+	cors := newPlatformGlobalCORS()
+	rr := httptest.NewRecorder()
+	handler := cors(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(200) }))
+
+	pc := &auth.ProjectContext{
+		ProjectID:  "p-fresh",
+		SchemaName: "tenant_fresh",
+		Slug:       "fresh",
+		AuthConfig: []byte("{}"), // never edited — ParseAuthConfig falls back to defaults
+	}
+	ctx := auth.ContextWithProject(context.Background(), pc)
+	handler.ServeHTTP(rr, makeReq(http.MethodOptions, "http://localhost:3000", ctx))
+
+	if rr.Header().Get("Access-Control-Allow-Origin") != "http://localhost:3000" {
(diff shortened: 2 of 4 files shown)


STGime (commented):
## Review — Approve

Fixes the exact silent-CORS-wall I hit on predwell. Verified:
- `DefaultAuthConfig()` now sets `CORSOrigins` = `RedirectURLs`, and the `TestDefaultAuthConfig_CORSMatchesRedirectURLs` test pins that they agree.
- Crucially, **no retroactive backfill**: `TestCORS_StoredConfigWithoutCORSOriginsIsNotBackfilled` confirms a project whose `auth_config` was edited keeps its explicit (empty) list — only never-configured projects (`{}`/null → `ParseAuthConfig` defaults) gain the localhost default. That's the right call; silently changing an existing project's CORS posture on upgrade would be worse than the bug.
- The `slog.Warn` per rejected origin (origin/project_id/path) closes the "invisible outside devtools" gap.

Security argument is correct: a remote page can't forge `Origin: http://localhost:3000`, and the SDK uses bearer tokens not ambient cookies, so a localhost CORS entry isn't an exposure.

One small note (not blocking): production projects created via the default now carry a `localhost:3000` CORS allowance indefinitely. Harmless per the above, but the `eurobase projects set-cors` follow-up you flagged would let operators trim it — worth keeping that on the list. LGTM.

PR #205, opened 2026-06-11
Fix #201: runner decrypts vault secrets with the right key_version

### LINKED ISSUE #201: ctx.vault.get returns null for existing secrets — runner missing/ stale VAULT_ENCRYPTION_KEY breaks all secret reads from edge functions
Found while wiring Mistral/Mollie into predwell edge functions.

## Symptom

A secret exists in the project vault (confirmed via `listSecrets` / MCP — `MISTRAL_API_KEY` present, correct name), but **every** edge function that calls `ctx.vault.get('MISTRAL_API_KEY')` gets `null`. Reproduced across multiple unrelated functions (parse-search, daily-digest) and it's the same reason `create-checkout` reports "payments not configured" for a set `MOLLIE_API_KEY`. So it's not function-specific — `ctx.vault.get` is returning null for any secret.

## Root cause (traced in functions-runner/vault.ts)

`resolveVaultSecret()` returns `null` at `if (!key)` when `VAULT_ENCRYPTION_KEY` isn't readable in the runner process:

```
const env = Deno.env.get("VAULT_ENCRYPTION_KEY") ?? "";
...
if (!env) { console.warn("[vault] VAULT_ENCRYPTION_KEY not configured; ctx.vault.get returns null"); return null; }
```

The gateway (Go, `cmd/gateway/main.go`) and the runner read the **same** env name `VAULT_ENCRYPTION_KEY`, and both pods use `envFrom: secretRef: eurobase-secrets` (`deploy/k8s/gateway.yaml`, `deploy/k8s/functions.yaml`). The gateway clearly has the key (it encrypts/stores secrets fine). So the functions pods should too — **unless they were started before the key was added to `eurobase-secrets`** (`envFrom` injects only at pod start; updating the Secret does not roll running pods). Other possibilities: a `key_version` mismatch vs migration 000057, or the key simply isn't in the Secret for the functions namespace/deployment.

## Impact

`ctx.vault.get` is unusable from edge functions → no function can read API keys for external services (LLMs, Stripe/Mollie, email). The whole "read a secret in your function" contract (closes #79) is silently broken in the deployed runner; it fails as `null` with no signal to the developer (the warning only goes to the pod's stdout, not `edge_function_logs`).

## Suggested fixes

1. **Verify + restart:** `kubectl exec deploy/functions -n eurobase -- printenv | grep VAULT_ENCRYPTION_KEY`. If absent, ensure it's in `eurobase-secrets` and `kubectl rollout restart deployment/functions -n eurobase`. If a Secret change predates the pods, this alone fixes it.
2. **Surface the failure:** when `ctx.vault.get` is called but `VAULT_ENCRYPTION_KEY` is unconfigured, record a one-line entry in `edge_function_logs` (or return a distinguishable error) so a developer can tell "secret missing" from "vault disabled in this runner" without pod access.
3. **Confirm `key_version` compat** between the gateway's encrypt path (000057) and the runner's single-key decrypt in `resolveVaultSecret`.

## Note: env_vars are NOT a vault substitute but currently work

`functions-runner/server.ts` reads `edge_functions.env_vars` as raw JSONB and passes it to `ctx.env` with no decryption (line ~400). So per-function `env_vars` are effectively plaintext-at-rest and bypass the vault path entirely — which contradicts the doc's "env_vars encrypted at rest (vault key)". Worth either encrypting them or fixing the doc.

STGime:
## Root cause

Migration 000057 (#179) moved the gateway to per-tenant HKDF-derived vault keys — every secret written or updated since is sealed at `key_version 1` — but the runner was left behind: `vault.ts` decrypts everything with the raw master key (the version-0 key), and `vault_get_for_runner` (000049) doesn't even return `key_version`. So **every post-000057 secret fails GCM auth in the runner** and `ctx.vault.get` returns `null`, indistinguishable from a missing secret. That's why `listSecrets`/console reads work (the gateway's own path is version-aware) while every edge function gets `null` (#201).

## Fix

- **Migration 000061**: `vault_get_for_runner` now also returns `key_version` (DROP + CREATE — return-type change — with 000049's owner/REVOKE/GRANT re-applied, wrapped in a transaction). An old runner selecting explicit columns is unaffected by the extra column; the new runner needs the migration first, which CI's migrate-before-deploy ordering guarantees.
- **`vault.ts`**: version-aware decryption mirroring `internal/vault/keyprovider.go` — version 0 uses the raw master; version ≥ 1 derives via WebCrypto HKDF-SHA256 with salt = tenant schema name and info = `eurobase-vault-v<version>`, plus a per-(version, schema) derived-key cache. The schema name comes from the HMAC-verified `X-Schema-Name` header — the same `projects.schema_name` value the gateway used as the HKDF salt.
- **Distinguishable failures** (issue fix 2): `resolveVaultSecret` returns `{value}` or `{error}`. Platform-side failures (key unconfigured, lookup failed, wrong key/decrypt failure) now surface to user code as a **thrown Error** with an actionable message, via the worker bridge's existing error path; a warning also lands in the function's log capture. A genuinely missing secret stays a plain `null`.

## Cross-language compatibility pinning

The exact derived key + sealed fixture (generated with Go `crypto/hkdf` + `crypto/cipher`, mirroring `keyprovider.go`) is asserted in **both** `internal/vault/keyprovider_test.go` (test-go job) and `functions-runner/vault_test.ts` (test-functions-runner job) — a drift in either implementation fails its CI job. WebCrypto HKDF was verified byte-identical to Go's `hkdf.Key` for this vector before embedding.

## Remaining items from the issue (not in this PR)

- **Ops check**: `kubectl exec deploy/functions -n eurobase -- printenv | grep VAULT_ENCRYPTION_KEY` to rule out a stale pod env as a second contributing cause (the key_version bug is deterministic regardless).
- **env_vars stored in plaintext** despite docs claiming "encrypted at rest (vault key)" — separate discrepancy, should get its own issue.

Closes #201

### DIFF
--- a/functions-runner/bridge.ts
+++ b/functions-runner/bridge.ts
@@ -53,8 +53,11 @@ export type ParentToWorker =
   // RPC result for `db.sql`. Mirrors the call's `id` so async handlers
   // route to the right Promise.
   | { type: "db.sql.result"; id: string; rows?: unknown; error?: string }
-  // RPC result for `vault.get`.
-  | { type: "vault.get.result"; id: string; value: string | null }
+  // RPC result for `vault.get`. `error` is set for platform-side
+  // failures (key unconfigured, lookup/decrypt failure, #201) and makes
+  // the worker's RPC layer throw; a missing secret is value: null with
+  // no error.
+  | { type: "vault.get.result"; id: string; value: string | null; error?: string }
   // RPC results for `ctx.storage.*` (closes #85).
   | { type: "storage.upload.result"; id: string; key?: string; size?: number; error?: string }
   | { type: "storage.signed_url.result"; id: string; url?: string; expiresAt?: string; error?: string }
--- a/functions-runner/server.ts
+++ b/functions-runner/server.ts
@@ -436,20 +436,29 @@ async function runUserHandlerInWorker(opts: {
         }
         case "vault.get.call": {
           // Closes #79: read the encrypted blob via the SECURITY DEFINER
-          // helper public.vault_get_for_runner (added by migration 000049)
-          // and decrypt locally with VAULT_ENCRYPTION_KEY. Failure modes
-          // (missing secret, no encryption key, decrypt failure) all
-          // resolve to null so the worker contract stays `string | null`.
-          resolveVaultSecret(db, projectId, msg.name)
-            .then((value) => {
+          // helper public.vault_get_for_runner (migration 000049,
+          // key_version added in 000061) and decrypt locally with the
+          // key for the version the secret was sealed with (#201).
+          // Missing secret → null; platform-side failures (no key,
+          // lookup/decrypt failure) → error, which the worker's RPC
+          // layer surfaces to user code as a thrown Error so "secret
+          // missing" and "vault broken" are distinguishable.
+          resolveVaultSecret(db, projectId, schemaName, msg.name)
+            .then((result) => {
               if (settled) return;
-              const reply: ParentToWorker = { type: "vault.get.result", id: msg.id, value };
+              if ("error" in result) {
+                logCapture.warn(`vault.get(${JSON.stringify(msg.name)}) failed: ${result.error}`);
+                const reply: ParentToWorker = { type: "vault.get.result", id: msg.id, value: null, error: result.error };
+                worker.postMessage(reply);
+                return;
+              }
+              const reply: ParentToWorker = { type: "vault.get.result", id: msg.id, value: result.value };
               worker.postMessage(reply);
             })
             .catch((err) => {
               if (settled) return;
               console.error(`[fn:${projectId}] vault.get failed`, err);
-              const reply: ParentToWorker = { type: "vault.get.result", id: msg.id, value: null };
+              const reply: ParentToWorker = { type: "vault.get.result", id: msg.id, value: null, error: "vault unavailable: internal error" };
               worker.postMessage(reply);
             });
           break;
--- a/functions-runner/vault.ts
+++ b/functions-runner/vault.ts
@@ -1,10 +1,22 @@
-// Vault read path for edge functions. Closes #79.
+// Vault read path for edge functions. Closes #79; key_version support
+// closes #201.
 //
 // Worker calls ctx.vault.get(name) → bridge.ts emits a vault.get.call →
 // server.ts (parent) calls resolveVaultSecret() here → we query
 // public.vault_get_for_runner(projectId, name) (SECURITY DEFINER, granted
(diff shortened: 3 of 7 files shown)


STGime (commented):
## Review — Approve (this unblocks predwell directly)

Sharper root cause than my issue guessed: not a stale pod, but that 000057 moved the gateway to per-tenant HKDF keys while the runner kept decrypting with the raw master — so every post-000057 secret fails GCM auth and returns `null`. That matches exactly what I'm seeing (`MISTRAL_API_KEY` set, `ctx.vault.get` → null, while `listSecrets` works).

**Verified the crypto crux independently:**
- Go `internal/vault/keyprovider.go`: `hkdf.Key(sha256.New, master, []byte(tenant), "eurobase-vault-v"+version, 32)` — full HKDF (extract+expand), salt = tenant/schema, versioned info.
- Deno `vault.ts`: WebCrypto `deriveBits({name:"HKDF", hash:"SHA-256", salt: schemaName, info: "eurobase-vault-v"+version}, …, 256)` — same salt, same info, 32 bytes. WebCrypto HKDF also does extract+expand, so it matches Go's `hkdf.Key` (it would NOT match a bare `Expand`).
- Both suites pin the **same vector** — derived key `RF884Rlm9sIYlm8ig2EQzav21+5drtmIg+vEujfkG54=` and the sealed fixture decrypting to `mistral_api_key_value` (`keyprovider_test.go` Go-side, `vault_test.ts` Deno-side). A drift in either implementation fails its CI job. This is the right way to guarantee cross-language seal/open compatibility.
- Migration 000061 DROP+CREATEs `vault_get_for_runner` to add `key_version`, re-applying the 000049 owner/REVOKE/GRANT; rollout-safe (old runner selects explicit columns; new runner needs the migration first, gated by the migrate-before-deploy job); down restores the 000049 signature.
- `key_version` missing → treated as 0 (legacy), tested; wrong-schema derive → GCM failure surfaced as error, tested.

**One behavior change worth calling out explicitly in the release notes:** `ctx.vault.get` now **throws** on platform-side failure (key unconfigured / lookup / decrypt) instead of returning `null`; a genuinely missing secret still returns `null`. That's the correct distinction and what I asked for — but any existing function that did `const k = await ctx.vault.get(x); if (!k) …` and relied on null-on-everything will now propagate a 500 on a misconfigured runtime. For predwell that's strictly better (I'd rather see the error than a silent 503). The worker bridge already rejects the RPC promise on `error`, so the throw path is wired. Verified end-to-end in the diff.

Strong LGTM — please ping me when it's deployed so I can confirm `parse-search` / `daily-digest` light up.

PR #207, opened 2026-06-11
CI: skip deploys whose commit is no longer the tip of main

STGime:
## Problem

The 2026-06-11 merge train (#202–#205) exposed a gap in the `deploy-production` concurrency group (#199): GitHub supersedes the *previously pending* job when a new one enters the group, but entry order is when each run's deploy job gets scheduled (tests + runner availability) — **not commit order**. The newest commit's deploy got cancelled and the oldest surviving run deployed last, leaving production on stale code without the newest migration until a manual `gh run rerun`.

## Fix

A freshness guard as the deploy job's first step — it runs **after** the concurrency queue admits the job, so it sees any commits that landed during the wait. If `github.sha` is no longer the tip of `main`, all 14 subsequent steps are skipped with a notice and the run ends green: an out-of-order survivor can no longer regress production.

Deliberately **not** auto-dispatching a deploy of the new tip from the stale run: that would need `actions: write` on the workflow token and risks duplicate image builds. Instead the notice names the recovery command (`gh run rerun <id> --failed`) for the case where the tip's own run was cancelled — production stays on the previous good deploy until then: safe and visible.

The Discord success notification is also gated so a skipped deploy doesn't announce itself as deployed.

## Verification

YAML validated; guard condition applied to all deploy steps (checkout → verify-rollout); the tip check uses `gh api repos/.../branches/main` with the default read-only token.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,13 @@ jobs:
       # concurrency group, nothing deploys it automatically — re-run that
       # run's deploy job (`gh run rerun <id> --failed`); production stays
       # on the previous good deploy until then.
+      #
+      # NOTE for future edits: every step added to this job MUST carry
+      #   if: steps.fresh.outputs.stale != 'true'
+      # A job-level gate (separate freshness job) would NOT work — it
+      # would run before this job enters the concurrency queue, missing
+      # commits that land during a 5-10 min wait behind another deploy.
+      # The per-step condition is the price of checking post-admission.
       - name: Check this commit is still the tip of main
         id: fresh
         env:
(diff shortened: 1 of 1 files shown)


STGime (commented):
## Review — deploy freshness guard

**Verdict: correct, well-reasoned fix.** The root-cause analysis is right: GitHub's concurrency supersession is by *queue-entry* order (test/runner timing), not commit order, so an older survivor can deploy last. Checking the tip *after* the queue admits the job is the correct place to catch it.

**Verified**
- `ci.yml` parses; the guard compares full SHAs (`gh api …branches/main --jq .commit.sha` vs `GITHUB_SHA`).
- All 14 deploy steps (checkout → verify-rollout) carry `if: steps.fresh.outputs.stale != 'true'`, and the **success** notification is additionally stale-gated so a skipped run can't announce a deploy.
- The **failure** notification is correctly *not* stale-gated: a stale run skips every step and ends green, so `failure()` is false and it won't misfire.
- Uses the default `contents: read` token (reading a branch needs no extra scope).

**Main suggestion (maintainability, non-blocking).** Gating 14 steps with a repeated per-step `if` is fragile — any deploy step added later silently runs on stale commits if someone forgets the guard. Cleaner: make the tip check its own tiny job and gate the whole deploy job once:
```yaml
freshness:
  runs-on: ubuntu-latest
  outputs: { stale: ${{ steps.c.outputs.stale }} }
  steps: [ ...the gh api check... ]
deploy:
  needs: [tests…, freshness]
  if: needs.freshness.outputs.stale != 'true'
```
One condition, future-proof, and the deploy job then shows as *skipped* rather than green-with-everything-skipped.

**Note on the residual failure mode (already acknowledged in the PR).** The guard prevents *regression*, but the newest commit can still end up **silently undeployed** if its own run was the one cancelled — and a skipped stale run ends **green**, so anyone watching for green checks won't notice the tip isn't live. The `::notice` helps but is easy to miss. Since `#deploys` is already wired, consider posting the "main moved, skipped — tip may be undeployed" case there too so the recovery prompt is visible. (Avoiding auto-dispatch is the right call given the `actions: write` cost.)

Safe to merge as-is; the job-split is a nice-to-have hardening.

---

STGime:
Adopted the Discord stale-skip notification (with the recovery command in the message) — agreed that a green skipped run is too quiet for "the tip may be undeployed."

On the job-split suggestion: I considered it but it would reopen the gap this PR closes. A separate `freshness` job necessarily completes **before** the deploy job enters the `deploy-production` concurrency queue — so it can't see commits that land while the deploy job waits in that queue behind another deploy (5–10 minutes, which is exactly the window from the 2026-06-11 incident pattern). The post-admission first-step check is the only placement that covers the queue-wait window; the repeated per-step `if` is the price. I've added a prominent NOTE comment above the guard so anyone adding a deploy step knows to carry the condition (and why the structure can't be "cleaned up" into a job gate).

PR #208, opened 2026-06-11
Add CLI release pipeline: versioned binaries + Homebrew tap

STGime:
## Problem

The console docs have advertised `brew install eurobase/tap/eurobase` "or download from GitHub releases" — but no CLI binary was ever published, no tap repo existed, and the CLI had no version command. Beta users had no way to install or update the CLI, and a stale local binary was undiagnosable — which is exactly how the pre-#192 broken `eurobase status` output resurfaced today after the fix had already shipped.

## What this adds

- **`eurobase version` / `--version`** — version + commit injected via ldflags; `dev` identifies from-source builds. Verified locally.
- **`.github/workflows/release-cli.yml`** — triggered by `cli-v*` tags (prefix keeps CLI releases distinct from the SDK's `v*` tags):
  1. Cross-compiles `cmd/eurobase` for darwin/linux × amd64/arm64 (CGO off, trimpath, stripped).
  2. Publishes tarballs + sha256 checksums as a GitHub release on the **public [STGime/homebrew-tap](https://github.com/STGime/homebrew-tap)** — the source repo is private, so assets must be publicly downloadable for users and brew; the tap doubles as the release host (one public repo total).
  3. Regenerates `Formula/eurobase.rb` from the real checksums via the GitHub contents API, so **`brew install stgime/tap/eurobase`** and `brew upgrade` track releases automatically.
- **Console docs** — install section updated to the real tap name and binary-download URL.

## Setup already done / remaining

- ✅ Public tap repo exists (`STGime/homebrew-tap`)
- ✅ `CLI_RELEASE_TOKEN` secret (fine-grained PAT, contents:write on the tap) configured
- After merge, cut the first release: `git tag cli-v0.3.0 && git push origin cli-v0.3.0`

## Verification

- `eurobase version` prints the ldflags-injected values (built + run locally).
- Formula generation simulated end-to-end with fixture checksums — output is valid Ruby (`ruby -c`) with correct URLs/shas.
- Workflow YAML validated.

Note: the tap lives under the personal account because the `eurobase` GitHub org name is taken by a third party. If that ever changes, retargeting is a one-line `TAP_REPO` edit + docs update.


STGime (commented):
## Review — CLI release pipeline

**Verdict: clean and working.**

**Verified locally**
- `go build -ldflags "-X main.version=… -X main.commit=…" ./cmd/eurobase` → both `eurobase version` and `eurobase --version` print the injected values; `fmt` is imported; `dev`/`none` defaults identify from-source builds.
- `release-cli.yml` parses; the unquoted `<<EOF` heredoc correctly expands `$(sha …)`/`${BASE}`/`${VERSION}` while leaving Ruby `#{bin}` literal; least-privilege `permissions: contents: read` with the tap PAT scoped separately is the right call.
- Docs updated to the real `stgime/tap/eurobase` + binary URL, consistent with the generated formula.

Publishing release assets to the public tap because the source repo is private is a sound workaround, and keeping it to one public repo is tidy.

**Minor notes (non-blocking)**
1. **`gh release create` isn't idempotent** — re-running the same `cli-v*` tag (or re-tagging after a failed run) errors because the release already exists. Consider `--clobber` on assets or a "delete release if it exists" guard so a partial failure is recoverable without manual cleanup.
2. **darwin binaries are unsigned / not notarized.** Fine through Homebrew (brew strips the `com.apple.quarantine` attr on install), but users who take the **manual download** path from the tap releases will hit Gatekeeper. Worth a one-line caveat (or `xattr -d com.apple.quarantine` hint) next to the binary-download URL in the docs.
3. The generated formula hardcodes `version`/URLs and is overwritten each release — intended "do not edit by hand" design; just flagging a manual tap edit would be clobbered (the header already says so).

Safe to merge.

---

Issue #191, opened 2026-06-10
DDL API gaps: composite UNIQUE, CHECK constraints, multi-column/partial indexes, array & parameterized types


STGime:
Found while building predwell on Eurobase. Enhancement request for `/v1/db/schema/tables/*` (`internal/query/ddl.go` / `ddl_handler.go`).

## Current limitations

- `POST .../constraints/unique` takes a single `column` — no composite UNIQUE. Natural keys like `(source, source_id)` on an aggregated listings table or `(district_id, date)` on a daily rollup can't be enforced; dedup moves into application code where it's race-prone.
- No CHECK constraints at all (`status IN (...)`, range checks) — validation is app-only.
- `POST .../indexes` is single-column, no partial (`WHERE is_active`) or composite indexes.
- `allowedTypes` rejects array types (`uuid[]`, `text[]`) and parameterized types (`numeric(8,2)`, `varchar(63)`); only bare type names pass `validateColumnType`.

## Suggested additions (kept within the validated-identifier security model)

- `columns: ["a","b"]` on the unique-constraint and index endpoints
- optional `where` (validated expression, same validator as RLS policy `using`) on index creation
- a `checks: [{name, expression}]` field on create-table plus a `POST .../constraints/check` endpoint, reusing the policy-expression validator
- extend `allowedTypes` with `<type>[]` arrays and `numeric(p,s)`/`varchar(n)` via a small grammar instead of the exact-match map

🤖 Filed with [Claude Code](https://claude.com/claude-code)

PR #209, opened 2026-06-11
Tenant-level schema migrations (#190): endpoint, CLI, docs

### LINKED ISSUE #190: No tenant-level migration channel — tenants must hand-roll schema tooling over the DDL REST API

## Problem

There is currently no sanctioned way for a tenant project to run versioned schema migrations:

- `POST /v1/db/sql` is SELECT-only (good security default, but rules out `.sql` migration files)
- MCP `runSQL`/`runSQLTransaction` are read-only unless the server is restarted with `EUROBASE_MCP_ALLOW_WRITES=true`
- `eurobase migrations up/down` operates on *platform* migrations "against a local/dev database", not on a tenant schema
- The DDL REST surface (`/v1/db/schema/tables/*`) works with the secret key but is imperative and unversioned

For predwell I ended up writing a bespoke runner (`apply-migrations.mjs` + an `app_migrations` bookkeeping table) that drives the DDL endpoints from `.mjs` migration files. Every serious tenant will have to reinvent this, and CLAUDE.md templates generated by `eurobase init` even tell agents to keep `migrations/NNNNNN_*.up.sql` files — which nothing on the platform can execute.

Side note: the obvious bookkeeping table name `schema_migrations` collides with the platform's own golang-migrate table, which is visible in tenant SQL scope (that visibility may itself be worth a look).

## Suggested fix

First-class tenant migrations, e.g.:

- `eurobase migrations up` (tenant mode): applies `migrations/*.sql` from the project directory against the tenant schema via a gateway endpoint that allows DDL+DML with the secret key, with version bookkeeping in a platform-managed table; or
- a `POST /v1/db/migrate` endpoint (secret key) accepting `{version, sql}` with the same semantics.

That would also fix the `eurobase init` CLAUDE.md template so its migration guidance matches something executable.

STGime:
## Problem

Tenants had no sanctioned way to run versioned schema migrations — `/v1/db/sql` is SELECT-only, the DDL REST surface is imperative and unversioned, and `eurobase migrations` was a platform tool. Every serious tenant (predwell included) had to hand-roll a runner over the DDL endpoints. Full design rationale: [#190 design comment](https://github.com/STGime/euroback/issues/190#issuecomment-4679992832).

## Gateway — one chokepoint, platform plane only

- `POST/GET /platform/projects/{id}/migrations` — apply `{version, name, sql}` / list history. **Platform auth (PAT/console) + developer role + migrator elevation on the developer pool** — the exact trust plane that already serves `/platform/.../schema/*`, so no new exposure class. Deliberately **not** on `/v1`: a leaked data-plane key must never escalate from data exposure to schema changes (the Supabase-derived plane-separation principle from the design comment).
- `ApplyTenantMigration`: one transaction per migration — per-project advisory lock (two CI runs can't race), `SET LOCAL ROLE` + `search_path` pinned to the tenant schema, 60s `statement_timeout`, user SQL via the **simple protocol** (multi-statement bodies are the point of a migration; this is the documented inverse of the cron-SQL hardening GHSA-fjjq-cqq9-q793, with containment from role/search_path pinning + the validator instead of single-statement mode), bookkeeping row in the same transaction.
- **Checksummed history**: identical re-apply is a no-op (idempotent `up`), re-applying a version with different SQL is rejected with 409 — applied migrations are immutable; add a new version.
- **Validator** (defense-in-depth, scans literal-stripped SQL): rejects references to other schemas (exact allowlist for `public.is_service_role/current_end_user_id/is_internal_auth_path/uuid_generate_v4` so real RLS policies work), `SECURITY DEFINER` (would permanently capture migrator privileges in a tenant-callable function), GRANT/REVOKE, SET/RESET role/search_path, transaction control, role/database/extension/schema management, ALTER SYSTEM, COPY. plpgsql function bodies and DO blocks stay allowed — the same accepted surface as the existing RPC-function endpoint.
- Migration `000062`: platform-managed `public.tenant_migrations` (also ends the tenant-side `schema_migrations` name collision noted in the issue). Applications land in the tamper-evident audit log as `schema.migration_applied`.

## CLI

- `eurobase migrations` is now the **tenant** command: `new <name>` scaffolds `migrations/NNNN_name.sql`, `up` applies pending in version order (idempotent, CI-safe), `status` shows local vs applied (including applied-elsewhere versions). Platform tooling moved to `eurobase admin migrations`.
- `eurobase login --token <PAT>` — non-interactive CI login (validates the token, runs the #192 project reconciliation).

## Docs (per the [documentation requirement](https://github.com/STGime/euroback/issues/190#issuecomment-4679992951))

- Console: new **"19. Schema Migrations"** section — workflow, realistic worked example (composite UNIQUE + CHECK + partial index + RLS policy: everything #191 says the DDL API can't express), execution semantics, GitHub Actions recipe with a PAT, and each validator boundary with its rationale.
- `eurobase init` CLAUDE.md template now documents the real, executable migration workflow — the template/reality mismatch was the original #190 complaint. `.cursorrules` no longer claims direct-PostgreSQL access exists.

## Tests

22 validator cases (realistic migrations accepted incl. all #191 shapes; escape hatches rejected; helper allowlist is prefix-attack-proof), checksum stability, literal-stripping scanner. `go build`/`vet`/`test` and the console build pass. The apply path itself needs a live DB — exercised post-deploy (predwell's hand-rolled runner is the natural first consumer).

## Relation to #191

Composite UNIQUE, CHECK, and partial/composite indexes are now expressible through migrations — #191 shrinks to "which of these still deserve first-class DDL-API support for agents."

Closes #190


STGime (commented):
## Review — tenant-level schema migrations

> Posting as a comment (GitHub blocks formal "Request changes" on one's own PR), but treat this as **blocking / do-not-merge**: there's a critical privilege-escalation / tenant-isolation hole in the containment model.

The feature itself is well-built (per-project advisory lock, immutable checksummed history, audit logging, `search_path` pin, `statement_timeout`, a solid literal-stripping scanner with good test coverage). But the core security control is bypassable — verified with a PoC against the actual validator.

### Critical: arbitrary SQL runs as `eurobase_migrator`, and the validator never sees function/DO bodies

`ApplyTenantMigration` does `SET LOCAL ROLE eurobase_migrator` (`engine.go:103`) — the role that **owns `public.*` and every tenant schema**. Containment against escaping the tenant schema rests on exactly two things, and they have overlapping gaps:

1. **`search_path` pin** — only governs *unqualified* name resolution. Explicitly qualified names (`public.projects`, `tenant_other.users`) ignore it.
2. **The regex validator** — `stripSQLLiterals` **blanks dollar-quoted bodies before scanning** (by design: "function bodies … are not scanned either"). So anything inside a `DO $$ … $$` block or a plpgsql function body is invisible to every denylist rule.

A project developer (platform plane, `developer` role) can put the payload in a function/DO body, which then runs as migrator. A throwaway test against `ValidateTenantMigrationSQL` — **all four pass validation**:

```
ACCEPTED: DO $$ BEGIN UPDATE public.projects SET plan='pro'; END $$;
ACCEPTED: CREATE FUNCTION pwn() ... LANGUAGE plpgsql AS $$ BEGIN UPDATE public.projects ... END $$; SELECT pwn();
ACCEPTED: DO $$ BEGIN EXECUTE 'GRANT eurobase_migrator TO eurobase_gateway'; END $$;
ACCEPTED: DO $$ ... SELECT * FROM tenant_other.users ... END $$;
```

Impact, all reachable by any project developer:
- **Cross-tenant breach** — read/write another tenant's schema (`tenant_other.users`); the primary isolation invariant, broken.
- **Platform-state tampering** — `UPDATE public.projects` (free self-upgrade to Pro; alter another project's owner/plan/billing).
- **Defeats this PR's own plane-separation goal** — `GRANT eurobase_migrator TO eurobase_gateway` makes the **data-plane** role migrator-equivalent. The stated rationale is "a leaked data-plane key must never escalate to schema changes"; this path lets a developer hand the gateway role migrator, after which a leaked secret key *does* have migrator.

The "no new exposure class / same as `/platform/.../schema/*`" framing doesn't hold: those endpoints run **structured, parameterized DDL helpers**, not arbitrary multi-statement SQL with executable bodies. Same caller plane, strictly greater capability. And extending the denylist into function bodies can't close this — `EXECUTE format(...)` dynamic SQL is unanalyzable statically.

### Fix (the real one): don't run migrations as `eurobase_migrator`

Run them under a **per-tenant, schema-scoped DDL role** — owns/`CREATE` only its own `tenant_<id>` schema, with **no** privileges on `public.*` tables or other tenant schemas — mirroring the existing `<schema>_func` / `eurobase_function_runner` pattern in `CLAUDE.md`. Then Postgres privilege enforcement contains qualified cross-schema and platform writes **regardless of what's inside a function body**, and the validator correctly becomes defense-in-depth instead of the sole control. (The RLS helpers `public.is_service_role()` etc. only need EXECUTE, which the per-tenant func roles already hold.)

### Minor (after the above)
- Denylist rules like `\bcopy\b` and `\bbegin\s*;` will false-positive on legitimate identifiers (e.g. a column named `copy`) once real migrations arrive — fine to defer, just flagging.
- Everything else (lock on `hashtextextended(projectID,190)`, checksum immutability → 409, no-op on identical re-apply, `schema.migration_applied` audit, 60s timeout) is well done and should stay.

Happy to re-review once migrations execute under a schema-scoped role.

---

STGime:
## Confirmed — blocking. Reproduced the bypass, fix in progress.

Ran all four payloads through the actual `ValidateTenantMigrationSQL`:

```
BYPASS (accepted): DO block updates public.projects
BYPASS (accepted): plpgsql function tampering
BYPASS (accepted): DO block grants migrator
BYPASS (accepted): DO block cross-tenant read
```

The analysis is exactly right: `stripSQLLiterals` blanks dollar-quoted bodies before the denylist runs, so a function/DO body is invisible to every rule — and even if it weren't, `EXECUTE format(...)` makes it statically unanalyzable. A regex validator can never be the containment boundary when the executing role is `eurobase_migrator` (owns `public.*` + every tenant schema). My "same plane as /schema/* = no new exposure" framing was wrong: those endpoints emit structured parameterized DDL, not arbitrary bodies — same caller, strictly greater capability.

**Marking this PR draft / do-not-merge.**

## Fix being implemented: migrations run under a per-tenant schema-scoped DDL role

New per-tenant role `tenant_<id>_ddl` (mirrors the existing `<schema>_func` pattern):
- `USAGE, CREATE` on its **own** tenant schema only; `USAGE` on `public` + `EXECUTE` on the RLS helpers (is_service_role / current_end_user_id / is_internal_auth_path / uuid_generate_v4) — and nothing else. **No** privileges on `public.*` tables, no other tenant schema, **not a member of `eurobase_migrator`**.
- The apply path does `SET LOCAL ROLE tenant_<id>_ddl` instead of migrator. Postgres then enforces containment regardless of what's inside a function/DO body or dynamic `EXECUTE`: `UPDATE public.projects` → permission denied; `tenant_other.users` → no USAGE; `GRANT eurobase_migrator …` → not an admin of that role. The validator stays purely as defense-in-depth / clearer error messages.

**Ownership convergence** (so the role can ALTER existing tables and the console DDL path keeps working): the ddl role owns application tables. `GRANT tenant_<id>_ddl TO eurobase_developer` lets the developer pool `SET ROLE` to it; `GRANT tenant_<id>_ddl TO eurobase_migrator` keeps the console DDL path (runs as migrator, now a member of the owner) able to manage them. `ALTER DEFAULT PRIVILEGES FOR ROLE tenant_<id>_ddl` grants DML on new tables to the gateway and `<schema>_func` roles so created tables are usable at runtime. A backfill reassigns ownership of existing application tables (everything in the schema except the platform-managed system tables) from migrator to the ddl role, and console `create-table` assigns new tables to the ddl role so both paths converge on one owner. Platform system tables (users, refresh_tokens, …) stay migrator-owned — correctly off-limits to tenant migrations.

Pushing shortly; the privilege enforcement gets a real-Postgres integration test in the style of `function_runner_role_test.go`.

STGime:
## Correction: the per-tenant role must be reached via a low-privilege LOGIN role, not the developer pool

While implementing the fix I found that "SET LOCAL ROLE tenant_<id>_ddl on the developer pool" — what I described above — is necessary but **not sufficient**. Second-order escape:

- The developer pool connects as `eurobase_developer`, which is an **INHERIT member of `eurobase_migrator`** (CLAUDE.md).
- `SET LOCAL ROLE tenant_<id>_ddl` can be undone from inside a migration body with `RESET ROLE` (in a `DO $$…$$` block or dynamic `EXECUTE` — invisible to the validator, same blind spot as the original finding). `RESET ROLE` returns to `session_user` = `eurobase_developer` → inherits migrator → `UPDATE public.projects` works again.
- The gateway pool is no better: `RESET ROLE` → `eurobase_gateway`, which holds `INSERT/UPDATE/DELETE on ALL TABLES IN public` (000037) — still writes `public.projects`.

So the session role that runs the migration must itself be harmless, so that `RESET ROLE` lands somewhere with no privileges. **This codebase already solved exactly this** for untrusted tenant SQL: `eurobase_function_runner` is a dedicated LOGIN role with no `public.*` grants and **NOINHERIT** membership of the per-tenant `<schema>_func` roles, so even `RESET ROLE` from inside user JS lands on a role with zero ambient privileges. Tenant migrations need the identical shape: a low-priv login role, NOINHERIT member of every `tenant_<id>_ddl`, used as the migration pool's connection role; `SET LOCAL ROLE tenant_<id>_ddl` per apply.

This turns the fix from a code change into a small **infra** change (a new login role created in the Scaleway console before its migration runs + a secret/pool wiring), so I'm holding implementation for a direction call rather than guessing at the role/secret setup. Options written up for the maintainer; the per-tenant `tenant_<id>_ddl` role model + ownership convergence (provision_tenant + backfill) is common to both and is ready.

STGime:
## Resolved — re-review please. Migrations now contained by a schema-scoped role, not the validator.

Pushed the fix. The containment boundary is now Postgres role privilege; the validator is demoted to defense-in-depth.

**What changed**
- **Migration 000063**: per-tenant `tenant_<id>_ddl` (NOLOGIN, `CREATE` on its own schema only, USAGE+EXECUTE on the public RLS helpers, no `public.*` table grants, **not a member of `eurobase_migrator`**) — owns the tenant's application tables. New login role **`eurobase_ddl_runner`** (LOGIN NOINHERIT, member of every `tenant_<id>_ddl`, no `public.*` beyond helper EXECUTE + SELECT/INSERT on `tenant_migrations`). provision_tenant + backfill create the per-tenant roles; ownership converges via migrator membership + `ALTER DEFAULT PRIVILEGES` + reassigning existing app tables (system tables stay migrator-owned).
- **Apply path** runs on the ddl-runner pool: bookkeeping as `ddl_runner`, user SQL under `SET LOCAL ROLE tenant_<id>_ddl`, then `RESET ROLE` before recording.
- **Gateway**: new `ddlRunnerPool` from `DATABASE_URL_DDL_RUNNER` (envFrom eurobase-secrets — no deployment YAML change). Endpoint **fails closed (503)** when unset; the route dropped `withDeveloperRole`.
- Console `create-table` assigns new tables to the per-tenant ddl role (savepoint-guarded) so console + migration DDL share one owner.

**Your four PoC payloads, now contained by the role** (not the regex):
- `UPDATE public.projects` in a DO body → permission denied (`tenant_<id>_ddl` has no grant on public.projects).
- `tenant_other.users` → no USAGE on that schema.
- `GRANT eurobase_migrator TO eurobase_gateway` → not an admin of that role.
- **`RESET ROLE` escape** (the second-order hole I found): lands on `eurobase_ddl_runner` (NOINHERIT, privilege-less) — `UPDATE public.projects` still denied. This is why it runs on a dedicated login role, not the developer pool.

Integration tests added (`tenant_ddl_role_test.go`, function-runner-test style — skip without a DB) asserting exactly those four denials plus a legitimate own-schema CREATE succeeding.

**Honest limitation**: I can't run a live Postgres here, so the role-privilege enforcement is verified by construction + the (skipped-locally) integration suite, not yet by execution. Worth a staging run once the role exists.

**Ops before deploy** (documented in CLAUDE.md, same convention as the other login roles): `CREATE ROLE eurobase_ddl_runner WITH LOGIN NOINHERIT` in the Scaleway console + add `DATABASE_URL_DDL_RUNNER` to `eurobase-secrets`. Safe to merge/deploy ahead of that — the feature stays dormant (503) until wired.

Also worth a fresh look: the `RELEASE/ROLLBACK TO SAVEPOINT` flow in `CreateTable`, and whether the migrator-membership-in-ddl-role grant has any unwanted reach.

STGime (commented):
## Recheck (commit `2722531`) — original finding fixed; a narrower cross-tenant hole remains

Big improvement, but I'd still treat this as **blocking**: a verified cross-tenant pivot survives.

### What's fixed ✅
The platform-escalation path is genuinely closed. Migrations now run as a per-tenant `tenant_<id>_ddl` (owns only its own schema) reached via the low-privilege `eurobase_ddl_runner` login role, which has **no `public.*` table grants**. So the three original impacts all now fail on privileges, even from inside a DO/function body: `UPDATE public.projects`, `GRANT eurobase_migrator TO eurobase_gateway`, and arbitrary platform writes. Fail-closed 503 when `DATABASE_URL_DDL_RUNNER` is unset, ownership convergence for the console DDL path, and `RESET ROLE` before bookkeeping are all well done.

### Residual (verified on Postgres 16): cross-tenant pivot via `RESET ROLE`
`eurobase_ddl_runner` is granted membership in **every** `tenant_<id>_ddl` (the backfill loop + `provision_tenant_ddl_role`), and `RESET ROLE` — reachable inside an unscanned DO/function body — returns to that shared login role, from which `SET ROLE` into *another* tenant's ddl role is permitted. I reproduced it with the exact 000063 role topology:

```
pinned_role = tenant_a_ddl            -- SET LOCAL ROLE by ApplyTenantMigration
-- migration body (invisible to the validator):
DO $$ BEGIN
  EXECUTE 'RESET ROLE';               -- -> eurobase_ddl_runner (shared login)
  EXECUTE 'SET ROLE tenant_b_ddl';    -- runner is a member of EVERY tenant ddl role
  EXECUTE 'DROP TABLE tenant_b.secrets';
END $$;
role_after_body = tenant_b_ddl
tenant_b_secrets_after = (null)        -- another tenant's table DROPPED
```

So a project developer can read/alter/drop **any other tenant's application tables**. Secondarily, after `RESET ROLE` the body runs as `eurobase_ddl_runner`, which holds `SELECT, INSERT ON public.tenant_migrations` — so it can also read every tenant's stored migration SQL and forge/pollute another project's history rows.

Narrower than before (no `public.*` platform tables, no migrator), but still a tenant-isolation breach — the core invariant.

**Root cause:** one shared login role that is a member of all per-tenant ddl roles, combined with `RESET ROLE` being reachable inside bodies. The `eurobase_function_runner` analogy doesn't transfer — edge functions can't issue raw `SET ROLE`/`RESET ROLE`, but the migration channel runs arbitrary SQL by design. Tightening the validator to scan bodies won't fix it either: `EXECUTE 'RE'||'SET ROLE'` evades static analysis (the same argument the PR makes for why the role boundary, not the validator, must be the control).

**Direction for a fix:** the connection's session/login role must be a member of **only the target tenant's** ddl role, so `RESET ROLE` can't pivot. Practically, a per-tenant `LOGIN` role (`tenant_<id>_ddl WITH LOGIN`, credential in the vault) that the gateway connects as per-apply — migrations are infrequent, so a short-lived per-migration connection is fine, and `RESET ROLE` then lands on the tenant's own harmless role. Any design where the session role can reach exactly one tenant works; the shared all-tenant-member runner is the specific problem.

Happy to recheck once the pivot is contained.

---

STGime:
## Confirmed again — blocking. The cross-tenant pivot is real; back to draft.

You're right, and I traced why the obvious patches don't close it:

- **Shared runner, member of all ddl roles** → `RESET ROLE` (reachable in an unscanned body) lands on the shared login role, `SET ROLE tenant_b_ddl` is permitted. Pivot. (Your repro.)
- **Dispatcher variant** (runner member of nothing; per-tenant SECURITY DEFINER `tenant_<id>.__apply(sql)` that the runner has EXECUTE on) — **same hole, isomorphic**: a body that `RESET ROLE`s to the runner can `SELECT tenant_b.__apply('DROP …')`, because the runner has EXECUTE on *every* tenant's dispatcher. "Member of all ddl roles" and "EXECUTE on all dispatchers" are the same problem.
- **NOINHERIT / validator-scans-bodies** — neither helps: SET ROLE membership isn't governed by INHERIT, and `EXECUTE 'RE'||'SET ROLE'` evades static analysis (the PR's own argument for why the role, not the regex, must be the boundary).

So the invariant the reviewer states is the only thing that works: **the session/login role must be able to reach exactly one tenant.** That means a per-tenant LOGIN identity the gateway connects as per-apply (`RESET ROLE` then lands on the tenant's own role, member of nothing → no pivot), plus session-user-bound bookkeeping (a SECURITY DEFINER recorder keyed off `session_user`, and no direct grant on `tenant_migrations` for the tenant role, so a body can't forge or read other projects' history — your secondary finding).

That's a real rework of the apply path (per-tenant connection + lazily-set per-tenant credential + bookkeeping helpers), and — importantly — it's the **third** subtle role-semantics issue in this channel that I can't execute against a live Postgres here to verify. Rather than ship a third unverified role design, I'm flagging this to the maintainer for a direction call: implement the per-tenant-login model now with a mandatory staging verification gate before merge, or decouple the safe foundation (validator/CLI/docs/role-model) from the execution path and land the apply endpoint only once it's verified live. Holding as draft either way.

STGime:
## Resolved — verified on real Postgres 16 this time. Re-review please.

You were right; I reproduced your pivot in a live container (tenant_a's migration dropped tenant_b's table), then built and verified the fix against real Postgres rather than by construction.

### Fix: per-tenant LOGIN roles the gateway connects as directly
The connecting/login role is now a member of **exactly one** tenant, so `RESET ROLE` in a body lands back on that same role and Postgres denies any `SET ROLE` into another tenant.

- **Migration 000063 (rewritten)**: `tenant_<id>_ddl` is NOLOGIN, **member of nothing**, owns its app tables. The shared `eurobase_ddl_runner` is gone. Bookkeeping via two `session_user`-bound SECURITY DEFINER helpers (`record_tenant_migration` / `tenant_migration_checksum`) — tenant roles have **no direct grant** on `public.tenant_migrations`, closing your secondary forge/read finding.
- **MigrationExecutor**: the developer pool (as migrator) sets the per-tenant role's LOGIN password (`HMAC-SHA256(DDL_PASSWORD_SECRET, schema)`, hex), then a short-lived connection **AS that role** runs the migration.
- Fails closed (503) without `DDL_PASSWORD_SECRET`; `DATABASE_URL_DDL_RUNNER` removed.

### Verified on Postgres 16 (reproducible)
`scripts/verify-tenant-migration-isolation.sh` (dependency-free, spins up Postgres, runs the battery) — all green:
- cross-tenant pivot via `RESET ROLE; SET ROLE tenant_b_ddl` → **permission denied**
- `UPDATE public.projects` → denied · cross-tenant read → denied · direct `tenant_migrations` forge/read → denied
- `tenant_b.secrets` intact · legitimate own-schema DDL (composite UNIQUE + partial index) succeeds
- bookkeeping bound to `session_user` (a body cannot forge another project's row)

I also caught three bugs only live testing surfaces (now fixed): the helper must strip the `_ddl` suffix for `session_user→schema`; `GRANT … WITH ADMIN OPTION` self-grant is rejected so migrator gets plain membership; migrator-as-creator (CREATEROLE) sets the password without ADMIN OPTION.

### Ops before deploy
Just add `DDL_PASSWORD_SECRET` (≥32 bytes) to `eurobase-secrets` — **no console role creation** (per-tenant roles are NOLOGIN, made by `provision_tenant`). Fails closed until set, so safe to merge/deploy ahead of it.

Worth a fresh look: the lazy `ALTER ROLE … PASSWORD` per apply (idempotent, hex literal), and the `MigrationExecutor` per-apply connection lifecycle.

STGime (commented):
## Re-review (commit `9a7ecd5`) — security objection resolved ✅

The move to **per-tenant `LOGIN` roles** closes the cross-tenant pivot. Verified on Postgres 16 with the new topology (a `tenant_<id>_ddl` LOGIN role that is a member of nothing, connected to directly):

- `RESET ROLE` → lands on `tenant_a_ddl` itself; `SET ROLE tenant_b_ddl` → **permission denied to set role**.
- Direct `SELECT`/`DROP` on `tenant_b.*` → **permission denied for schema tenant_b**.
- No `public.*` table grants → platform-table writes still denied (original finding stays closed).
- Bookkeeping via `record_tenant_migration` / `tenant_migration_checksum` — SECURITY DEFINER, owned by migrator, **scoped through `session_user`** to the caller's own project, with no direct grant on `public.tenant_migrations`, so a body can neither forge nor read another tenant's history.
- The old shared `eurobase_ddl_runner` (member-of-all, the thing that enabled the pivot) is fully removed; fail-closed 503 when `DDL_PASSWORD_SECRET` is unset.

Both the original migrator/platform escalation and the cross-tenant `RESET ROLE` pivot are resolved. This is the right architecture — nice iteration. Build + validator/DDL tests pass.

## One likely functional break + 2 hardening notes (not security-blocking)

1. **Likely breaks at runtime — missing `GRANT CONNECT`.** `tenant_<id>_ddl` is now a LOGIN role the gateway connects as directly, but `provision_tenant_ddl_role` never runs `GRANT CONNECT ON DATABASE eurobase TO <role>`. Every other login role gets one explicitly (`000037` gateway, `000044` developer, `000047` function_runner), which strongly implies PUBLIC's default CONNECT is revoked on this database. If so, **every migration apply fails at connect time** (`permission denied for database eurobase`). It only surfaces against a live DB (the apply path is "exercised post-deploy"), so the test suite won't catch it. One-line fix in the provisioning helper + the backfill loop.

2. **(Hardening) The role is left `LOGIN` with a deterministic password permanently** after first apply (`password = HMAC(DDL_PASSWORD_SECRET, schema)`), so a leaked secret makes every tenant DDL role loginable. Mitigated by the DB not being tenant-exposed, but demoting back to `NOLOGIN` after each apply (or rotating) would shrink the window.

3. **(Hardening) `ALTER ROLE … PASSWORD '<hex>'` is logged verbatim** under `log_statement = ddl`/`mod` — derived passwords could land in the Postgres server log. Worth checking the managed DB's logging level.

Net: the security objection is resolved and verified. I'd just confirm/fix the `CONNECT` grant (#1) before relying on it, since it likely makes the happy path fail in production.

---

STGime:
## Re-review items addressed — all verified on Postgres 16 (commit b488566)

Thanks for the approval on the architecture. All three follow-ups handled and re-verified against a live DB.

### 1. Missing GRANT CONNECT (the functional bug) — fixed
You were right. Reproduced on a prod-like DB (`REVOKE CONNECT ON DATABASE eurobase FROM PUBLIC`): the per-tenant LOGIN role's connection failed with `FATAL: permission denied for database "eurobase"`. `provision_tenant_ddl_role` now `GRANT CONNECT ON DATABASE eurobase TO <role>` (new tenants + the backfill loop), matching every other login role — confirmed the connect then succeeds. As with `000037`'s gateway CONNECT grant, this runs as migrator, which already grants CONNECT on this DB, so it works under the migrate job.

### 2 & 3. Hardening — addressed together
The role is now **demoted back to NOLOGIN after each apply**, so it is loginable only during the brief apply window. The executor does promote (`ALTER ROLE … LOGIN PASSWORD`) → connect → run → demote (`ALTER ROLE … NOLOGIN`, deferred on an uncancellable context), with a one-shot connect retry covering the race where a concurrent apply demoted between our promote and connect. This directly shrinks the #2 window, and **mitigates #3**: even if the `ALTER ROLE … PASSWORD` literal lands in the server log at `log_statement=mod/ddl`, a logged password is useless while the role sits NOLOGIN between applies.

### Verification (reproducible)
`scripts/verify-tenant-migration-isolation.sh` extended — PUBLIC CONNECT revoked, GRANT CONNECT added, ddl roles created *as migrator* (so migrator holds admin to promote/demote — another fidelity detail live testing surfaced), plus a new check asserting the role is NOLOGIN after demote and loginable after re-promote. **All 8 checks green** (pivot, platform write, bookkeeping forge/read, cross-tenant read, tenant_b intact, own-schema DDL, session_user-bound bookkeeping, promote/demote lifecycle).

Ops before deploy is unchanged: add `DDL_PASSWORD_SECRET` (≥32 bytes) to `eurobase-secrets`; fails closed (503) until then.

STGime (commented):
## Re-review (commit `b488566`) — prior items addressed; one quiet trap remains

Both points from the last round are handled, and the security design is unchanged (the `NOLOGIN` demote actually tightens it). Build + validator/DDL tests pass.

✅ **Lifecycle hardening is good.** The role is promoted to `LOGIN` only for the apply window, then demoted to `NOLOGIN` (defer uses `context.WithoutCancel` so a cancelled request still locks down), with a one-shot retry for the concurrent-demote race. This also neutralizes the password-in-DDL-log concern — a logged password is useless while the role is `NOLOGIN`.

### ⚠️ The `GRANT CONNECT` fix likely no-ops in prod — and fails *silently*

`provision_tenant_ddl_role` is `SECURITY DEFINER` owned by `eurobase_migrator`, so its `GRANT CONNECT ON DATABASE eurobase` runs **as migrator**. Granting CONNECT requires being the **database owner** (or holding `CONNECT … WITH GRANT OPTION`). If migrator is neither — plausible: CLAUDE.md only says migrator owns `public.*` and tenant schemas, not the DB, and the fact that every other login role needed an *explicit* CONNECT grant points to a bootstrap owner having done those — then the grant is a **no-op that emits only a `WARNING`, not an error**. Confirmed on Postgres 16:

```
SET ROLE eurobase_migrator;            -- CREATEROLE, not DB owner
GRANT CONNECT ON DATABASE postgres TO victim;
-- WARNING:  no privileges were granted for "postgres"
-- GRANT      (statement "succeeds" — the migration does NOT fail)
```

So `000063` would apply cleanly, the per-tenant role never actually gets CONNECT, and **every migration apply still fails at connect time** — the very thing this commit set out to fix, now masked because the grant warns instead of erroring. The new `verify-tenant-migration-isolation.sh` can't catch it either: it runs the `GRANT CONNECT` **as superuser** ("here as superuser to skip db-ownership"), so it passes regardless of whether the prod path works.

**Recommend one of:**
- Confirm `eurobase_migrator` owns the `eurobase` database (or grant it `CONNECT … WITH GRANT OPTION` once via the bootstrap owner) — the same identity question the `000037/000044/000047` CONNECT grants implicitly relied on; and/or
- Make it fail loud at migrate time rather than silent at first apply — assert the grant took:
  ```sql
  IF NOT has_database_privilege(v_ddl_role, 'eurobase', 'CONNECT') THEN
      RAISE EXCEPTION 'ddl role % lacks CONNECT — migrator cannot grant it (not db owner?)', v_ddl_role;
  END IF;
  ```

Net: the security objection stays resolved and the lifecycle work is solid; the remaining risk is this one operational gap, which is easy to make non-silent.

---

STGime:
## Quiet trap fixed — verified on Postgres 16 (commit 93a72e8)

Great catch. Reproduced it exactly: `GRANT CONNECT ON DATABASE eurobase` as a non-owner `eurobase_migrator` emits `WARNING: no privileges were granted` but the statement **succeeds**, so `000063` applies cleanly while the per-tenant role silently lacks CONNECT — and every apply then fails at connect time, masked.

**Fix — fail loud at the apply path, where it's actionable:**
- `MigrationExecutor` (as migrator) re-grants CONNECT on promote and verifies it with `has_database_privilege(role, db, 'CONNECT')` — true iff the role can actually connect (via the grant **or** PUBLIC), so it's the precise detector. On false, `Apply` returns a clear error naming the ops fix instead of an opaque "permission denied for database" at connect time. (Adopted your `has_database_privilege` suggestion, moved to apply-time so it doesn't hard-block the security role-model migration for operators not yet using migrations.)
- The migration keeps `GRANT CONNECT` as best-effort with a comment pointing at the apply-time check; `000063` still installs the role model where migrator can't grant CONNECT, the feature just stays loudly unusable until the ops fix.
- CLAUDE.md now states the requirement: **`eurobase_migrator` must own the `eurobase` database** (or hold `CONNECT … WITH GRANT OPTION`) — the same identity question the `000037/000044/000047` CONNECT grants implicitly relied on.

**Verification (reproducible):** `scripts/verify-tenant-migration-isolation.sh` gains a check that exercises the prod path — a non-owner migrator's `GRANT CONNECT` no-ops (`has_database_privilege` stays false: the trap), and making migrator the db owner makes it take. The script's earlier `GRANT CONNECT … as superuser` could indeed mask this; the new check connects as migrator-non-owner specifically. **All 9 checks green.**

STGime (commented):
## Re-review (commit `93a72e8`) — all findings resolved ✅

The `CONNECT` no-op trap is closed. On promote, `setRoleLogin` now `GRANT`s CONNECT and then verifies it with `SELECT has_database_privilege(role, db, 'CONNECT')`, failing loud with an ops-actionable message instead of letting a non-owner `GRANT` silently no-op into an opaque connect error. Verified on Postgres 16: non-owner migrator → `has_database_privilege` = `f` (raises the clear error); after `ALTER DATABASE … OWNER TO eurobase_migrator` → `t` (proceeds). The same-transaction visibility the guard relies on holds. Build + validator/DDL tests pass, and the isolation script now exercises the non-owner trap explicitly.

### Thread wrap-up — every objection across the rounds is addressed
- **Critical: arbitrary SQL ran as `eurobase_migrator`** (platform-table writes, `GRANT eurobase_migrator TO eurobase_gateway`) → fixed: per-tenant schema-scoped role, no `public.*` table grants.
- **Cross-tenant `RESET ROLE` pivot** → fixed and verified: per-tenant **LOGIN** role that is a member of nothing; `RESET ROLE` lands on itself, `SET ROLE`/cross-schema access denied.
- **Bookkeeping** → `session_user`-bound SECURITY DEFINER helpers scoped to the caller's own project; no direct grant on `public.tenant_migrations`.
- **Role lifecycle** → `NOLOGIN` except during the apply window (with the concurrent-demote retry), which also neutralizes the `ALTER ROLE … PASSWORD` log-exposure concern.
- **`CONNECT` silent no-op** → re-granted + verified at apply time; fails loud if migrator can't grant it.

No remaining code blockers — LGTM from my side (GitHub won't let me formally Approve my own PR, so this is a comment).

**One deployment-checklist item** (not a code defect): in the live environment, `eurobase_migrator` must own the `eurobase` database (or hold `CONNECT … WITH GRANT OPTION`). If it doesn't, the apply path now fails with the exact instruction rather than an opaque error — but it's worth doing as part of the cutover so the first tenant migration works.

---

PR #215, opened 2026-06-12
Fix #214: forward caller headers + query string to edge functions

### LINKED ISSUE #214: Gateway strips custom request headers + query string when proxying to edge functions
Found while building an OpenImmo intake endpoint (a webhook-style edge function) on predwell.

## Symptom

A `verify_jwt=false` function cannot read **any** custom request header or the **query string** of the invoke — both are dropped before the runner sees the request. A probe function returning `req.url` + all headers shows:

```
url:  "http://functions:8000/invoke"      ← original query string GONE (?key=… stripped)
headers: accept-encoding, content-type, host, transfer-encoding, user-agent,
         x-eurobase-signature, x-eurobase-timestamp, x-function-id,
         x-function-name, x-function-version, x-plan, x-project-id, x-schema-name
```

Sent on the request but **absent** in the runner: `X-Feed-Key`, `X-Custom-Thing`, and `?key=…&foo=bar`. Only the **body** and `Content-Type` survive. (`Authorization` can't be repurposed either — the gateway validates it as an end-user JWT and 401s an arbitrary value.)

## Why it matters

Edge functions are the natural place to build **webhook receivers and small HTTP APIs** (payment webhooks, OpenImmo/feed intake, partner callbacks, signed callbacks from third parties). Almost all of these authenticate or route via a **custom header** (`X-Signature`, `X-Api-Key`, `Authorization`) or a **query/path** (`?token=`). With both stripped, the only channel left is the request body — which doesn't work when the body format is fixed by the sender (raw XML, a provider's webhook JSON shape, form-encoded, etc.). I had to require partners to embed the key *inside* the OpenImmo XML (`<uebertragung benutzer="…">`) as a workaround — non-standard and brittle.

(Note: Mollie webhooks happen to work only because Mollie puts the payment id in the form body and we re-fetch from Mollie's API for trust — i.e. we got lucky that the auth-relevant data was in the body.)

## Suggested fix

Forward a safe allowlist of request headers to the runner (e.g. everything except hop-by-hop + the gateway's own `X-Eurobase-*`/`X-Function-*`), and **preserve the original query string** on the proxied URL so `new URL(req.url).searchParams` works in the function. If header forwarding is a concern, at minimum forward the query string and a small set of common auth headers (`X-Api-Key`, `X-Signature`, `X-Webhook-*`). Document exactly which headers a function can rely on.

STGime:
## Problem

The gateway proxied edge-function invocations to the runner with `runnerURL+"/invoke"` (query string dropped) and only its own headers set — so a `verify_jwt=false` function could read **neither** custom request headers (`X-Signature`, `X-Api-Key`, `X-Webhook-*`) **nor** the query string. That makes edge functions unusable as webhook receivers / small HTTP APIs unless the auth-relevant data happens to be in the body (why Mollie webhooks worked but OpenImmo/feed intake didn't).

## Fix

`HandleInvoke` now:
- **Preserves the original query string** on the proxied URL, so `new URL(req.url).searchParams` works in the function.
- **Forwards the caller's request headers** (`copyForwardableHeaders`), skipping three families:
  - hop-by-hop headers,
  - platform auth (`Authorization` / `apikey` / `Cookie` — these authenticate the call *to* Eurobase, not the function),
  - the gateway control namespace (`X-Eurobase-*`, `X-Project-*`, `X-Schema-Name`, `X-Plan`, `X-Function-*`, `X-User-*`, `X-Request-ID`).

## Security / protocol reasoning

The control namespace is denied from caller forwarding so a caller can't spoof a signed identity header — and a forwarded `X-Request-ID` would otherwise **break HMAC verification** (the gateway signs it empty; `canonicalMessage` covers it). The gateway sets its control headers **after** copying caller headers, then signs; the HMAC covers only that fixed set, so forwarding arbitrary extras is safe. The runner already passes `req.url` + all headers through to the worker — no runner change needed.

## Tests / docs

- `invoke_test.go`: forward/strip classification — custom headers forwarded (incl. multi-valued), platform-auth + control + hop-by-hop stripped, and the `isGatewayControlHeader` boundary.
- Console docs gain a "Reading the request" section documenting exactly which headers/query a function can rely on and the two withheld families (use `ctx.user` for end-user identity).

Closes #214


STGime (commented):
## Review — forward caller headers + query string to edge functions

**Solid fix, one credential-leak finding to address before merge.** The header classification is careful and the HMAC reasoning checks out — but the query-string change reopens, via a different path, the exact platform-auth leak the header strip prevents.

### What's right
- **HMAC invariant holds.** Checked `canonicalMessage` against the strip set: it signs exactly `X-Project-ID`, `X-Schema-Name`, `X-Function-ID`, `X-User-ID`, `X-User-Email`, `X-Plan`, `X-Request-ID`, plus `X-Eurobase-Timestamp/Signature` — and **every one** is denied to callers via `isGatewayControlHeader` / `unforwardableHeaders`. So forwarding arbitrary extra headers can't break verification or spoof a signed identity, and control headers are `Set` after the copy so they always win. The `X-Request-ID`-signed-empty reasoning is accurate.
- **Header classification is sound** — standard RFC 7230 hop-by-hop set; platform auth (`authorization`/`apikey`/`cookie`) stripped for confidentiality; multi-valued headers preserved. Good test coverage of forward/strip/boundary.
- Query-string preservation and `Content-Type`-via-copy are correct; runner needs no change.

### Finding (should fix before merge): `?apikey=` leaks into the function
`apikey_middleware.go:61-64` accepts the API key from the `apikey` header **or** `r.URL.Query().Get("apikey")`. This PR strips the `apikey` *header* (good) but forwards the **entire raw query string** unfiltered:

```go
if r.URL.RawQuery != "" {
    target += "?" + r.URL.RawQuery
}
```

So invoking `POST /v1/functions/foo?apikey=eb_sk_live_…` forwards the caller's secret key into the function's `req.url` — readable via `new URL(req.url).searchParams` and very likely to land in function logs or be forwarded onward. That violates the PR's own stated principle ("platform auth … must not leak to user function code"): the header path enforces it, the query path (which this PR just made readable) does not. It's most relevant for the exact case this PR enables — `verify_jwt=false` webhook/API functions invoked by URL.

**Fix:** strip auth-bearing params from the forwarded query, mirroring the header strip:
```go
if r.URL.RawQuery != "" {
    q := r.URL.Query()
    q.Del("apikey")          // consumed by the gateway, not the function
    if enc := q.Encode(); enc != "" {
        target += "?" + enc
    }
}
```
(`q.Encode()` re-sorts/re-encodes the query — fine for `searchParams`, just noting the wire form changes. A test asserting `apikey` is dropped from the forwarded URL would lock it in.)

### Minor (non-blocking)
- `copyForwardableHeaders` doesn't honor headers *named in* the `Connection` header (RFC 7230 says a proxy should drop those too). Harmless here since the runner is internal and it only over-forwards, but worth a comment.

Otherwise a clean, well-tested fix. With the `apikey` query strip added I'd be happy to approve.

---

PR #216, opened 2026-06-12
ops: tenant-migration enablement + recovery tooling and runbook

STGime:
## Operational tooling for tenant migrations (#190)

In-cluster one-off Job runners (the prod DB is reachable only from the cluster) plus a runbook, built while deploying and operating the tenant-migrations feature. These were used to recover the three `000063` deploy failures and to drive enablement.

### Scripts (`scripts/ops/`)
- **`migrate-status.sh`** — print the prod migration version + dirty flag.
- **`migrate-force.sh`** — clear a dirty `schema_migrations` flag by forcing the last-good version (default 62). Runs no SQL / changes no schema; prompts to confirm.
- **`set-ddl-password-secret.sh`** — step 1 of enablement: generate `DDL_PASSWORD_SECRET`, patch `eurobase-secrets`, roll the gateway. (Already run in prod — gateway logs "tenant migrations enabled".)
- **`grant-migrator-connect.sh`** — attempt step 2 (the CONNECT grant) as a DB admin; passes `PG*` env vars so an email-style username / special-char password don't break a connection URL. (In prod this needs `_rdb_superadmin`, so it ended up being a Scaleway support request — see runbook.)
- **`db-checks.sh`** — read-only state (db owner, migrator CONNECT grant option, ddl-role count) to confirm step 2 landed.

All pass the admin password via a hidden prompt + short-lived Secret, never on the command line; Jobs and temp secrets are cleaned up.

### Runbook (`docs/runbooks/enable-tenant-migrations.md`)
The two enablement steps, the copy-paste Scaleway support ticket for the `_rdb_superadmin` CONNECT grant, end-to-end verification, and the dirty-migration recovery flow.

### Why step 2 needs Scaleway
The per-tenant `tenant_<id>_ddl` roles are SQL-created by `provision_tenant_ddl_role`, so they don't get Scaleway's automatic CONNECT grant (only console/API-created users do). `eurobase_migrator` can't forward CONNECT (has it, no grant option) and the `eurobase` DB is owned by `_rdb_superadmin`, which customers can't log in as — so one grant must be run by Scaleway. Documented in the runbook; tracked.

### DIFF
--- a/docs/runbooks/enable-tenant-migrations.md
+++ b/docs/runbooks/enable-tenant-migrations.md
@@ -1,0 +1,85 @@
+# Enabling tenant migrations (#190) in production
+
+Tenant migrations (`POST /platform/projects/{id}/migrations`, `eurobase migrations up`)
+ship dormant and **fail closed (503)** until two one-time ops steps are done.
+Both are safe to do independently; the feature only works once both are complete.
+
+## Step 1 — `DDL_PASSWORD_SECRET` (done via script)
+
+The gateway derives each per-tenant ddl role's login password from this secret.
+
+```
+./scripts/ops/set-ddl-password-secret.sh
+```
+
+Generates a 32-byte secret, patches `eurobase-secrets`, and restarts the gateway.
+Confirm with the gateway log line `"tenant migrations enabled"`.
+
+## Step 2 — let `eurobase_migrator` grant CONNECT (needs `_rdb_superadmin`)
+
+The per-tenant `tenant_<id>_ddl` roles are LOGIN roles the gateway connects as.
+They are created by the SQL migration `provision_tenant_ddl_role`, so — unlike the
+runtime roles created through the Scaleway console — they do **not** get Scaleway's
+automatic CONNECT grant. `eurobase_migrator` cannot grant it onward (it has CONNECT
+but no grant option), and the `eurobase` database is owned by `_rdb_superadmin`,
+which customers cannot log in as or `SET ROLE` to.
+
+So one statement must be run **as `_rdb_superadmin`** — which on Scaleway managed
+Postgres means a support request.
+
+### Scaleway support ticket (copy-paste)
+
+> Subject: Run a one-time GRANT as the database superadmin on our PostgreSQL instance
+>
+> Instance: **eurobase-db** (`62deb4e1-4dba-44ea-8fbc-2d519637425a`), region **fr-par**, PostgreSQL 16.
+> Database: **eurobase**.
+>
+> Please run the following as the database owner / `_rdb_superadmin`:
+>
+> ```sql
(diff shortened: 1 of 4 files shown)


STGime (commented):
## Review — tenant-migration ops tooling + runbook

**Good ops hygiene; one real bug in `grant-migrator-connect.sh`.** All scripts use `set -euo pipefail`, hidden password prompts, short-lived secrets passed via `envFrom` (never on the CLI / in history), in-cluster Jobs with `ttlSecondsAfterFinished`, and `trap` cleanup. The runbook is clear and the Scaleway support-ticket text is genuinely useful. `migrate-status.sh` / `migrate-force.sh` referenced by the runbook already exist on `main`, so no dangling references. `mergeable_state: clean`.

### Finding — `grant-migrator-connect.sh` reports false success
It prints `✅ Granted` based on the **Job phase**, but a `GRANT CONNECT … WITH GRANT OPTION` run by a non-owner is a `WARNING`, not an error — so `psql -v ON_ERROR_STOP=1` still exits 0, the Job `Succeeds`, and the grant didn't take. The script even captures `migrator_canconnectgrant=false` in the output, then prints ✅ anyway. This is the same silent-no-op behavior #217 hit. Gate success on the verification result — easiest is to make the Job itself assert it:

```sh
psql -v ON_ERROR_STOP=1 \
  -c "GRANT CONNECT ON DATABASE eurobase TO eurobase_migrator WITH GRANT OPTION" \
  -c "DO \$\$ BEGIN IF NOT has_database_privilege('eurobase_migrator','eurobase','CONNECT WITH GRANT OPTION') THEN RAISE EXCEPTION 'grant did not take — run as _rdb_superadmin'; END IF; END \$\$;"
```
Then `phase=Succeeded` is meaningful. (Impact is limited since the runbook already routes operators to Scaleway support for this, but the ✅-on-no-op is misleading.)

### Minor
- The temp admin-cred secret lives briefly in the `eurobase` namespace (fine — operator-run, trap-cleaned); worth a one-line note that an operator with secret-create rights is assumed.

Otherwise solid.

---

STGime (commented):
## Re-review — finding resolved ✅

`grant-migrator-connect.sh` no longer false-succeeds. The Job now runs a `DO $$ … RAISE EXCEPTION $$` assertion on `has_database_privilege('eurobase_migrator','eurobase','CONNECT WITH GRANT OPTION')`, so a non-owner `GRANT` (which is a `WARNING`, not an error) flips the Job to `phase=Failed` instead of printing `✅ Granted` on a silent no-op — `phase=Succeeded` is now meaningful. Matches the #217 lesson. `mergeable_state: clean`.

No remaining issues — LGTM (comment, since it's my own PR).

---

PR #116, opened 2026-05-13
Security: mitigate CVE-2026-43284 (Dirty Frag) on Kapsule nodes

STGime:
Scaleway advisory 2026-05-12: Linux kernel LPE via the ESP IPsec transform modules (`esp4`/`esp6`) and the AF_RXRPC socket family (`rxrpc`). Same threat shape as CVE-2026-31431 (Copy Fail) — local unprivileged user can escalate to root or escape a container. The functions-runner pods execute untrusted tenant JS on shared Kapsule nodes, so the LPE → escape path is in scope.

## What lands

- **`deploy/k8s/security/disable-dirty-frag.yaml`** — DaemonSet mirror of the existing `disable-algif-aead.yaml`. Writes `install esp4 / esp6 / rxrpc /bin/false` into `/etc/modprobe.d/` on every node and `rmmod`s any already-loaded modules. Tolerates all taints so the autoscaler-brought-up nodes are also hardened on first boot.
- **`.github/workflows/ci.yml`** — one extra `kubectl apply -f` line in the security apply block (before any pod restart, same rationale as Copy Fail).

## Risk check

None of `esp4`, `esp6`, `rxrpc` are on the gateway / worker / functions-runner hot path:
- `esp4` / `esp6` are IPsec ESP transforms — we terminate TLS at the LB, not IPsec.
- `rxrpc` is the AF_RXRPC socket family (Linux AFS). No code in this tree opens AF_RXRPC sockets; pgx, gorilla/websocket, Deno, V8 crypto don't touch it.

Rollback: \`kubectl delete -f deploy/k8s/security/disable-dirty-frag.yaml\`, then \`modprobe esp4 esp6 rxrpc\` on each node.

## Operator follow-up (not in this PR)

The proper fix is a patched kernel. Roll the Kapsule node pool through the Scaleway console once the patched image is selected:

\`\`\`
scw k8s pool upgrade <pool-id>
\`\`\`

Rolling drain — \`kubectl rollout status\` gates in CI cover availability. After the roll, both DaemonSets self-apply to the fresh nodes; verify with \`kubectl get ds -n kube-system disable-algif-aead disable-dirty-frag\`.

This same Kapsule node-pool roll also closes the parallel **CrackArmor** AppArmor advisory (kernel patches only).

## Test plan

- [ ] Merge + deploy
- [ ] \`kubectl get ds -n kube-system disable-dirty-frag\` shows DESIRED == CURRENT == READY across all nodes
[...]

PR #220, opened 2026-06-25
ops: TLS cert monitor + runbook (post-incident 2026-06-24)

STGime:
Operationalises the "this should NEVER EVER happen" promise from the 2026-06-24 wildcard-cert outage.

## Files
- `scripts/ops/cert-status.sh` — exit-code-based per-host TLS freshness check (`openssl s_client` + `openssl x509 -checkend`). Runnable from a laptop, CI, and the CronJob below. Exit codes: `0` all valid ≥14d, `1` warn, `2` critical.
- `deploy/k8s/cert-monitor-cronjob.yaml` — hourly probe; warn → Discord embed; critical → embed with `@everyone`. Checks the **served** certificate, not cert-manager internal status (the June 24 outage had `Certificate.status[Ready]=True` while the Secret was stale).
- `docs/runbooks/tls-cert-expiry.md` — incident-of-record recovery flow, including the kubectl describe failure-class lookup table.
- `deploy/k8s/ingress.yaml` — wildcard `Certificate` points at `letsencrypt-prod-dns-cloudflare` (native cert-manager Cloudflare DNS-01 solver — already wired in the cluster). The Scaleway issuer is kept declared but dormant so the future DNS migration is a one-line `issuerRef` swap.
- `docs/emails/2026-06-24-cert-outage.html` — beta-user apology email about the outage; lives next to the runbook so post-incident comms have the same history as the fix.

## Validation
- `kubectl --dry-run=client apply` clean on both manifests.
- `scripts/ops/cert-status.sh` smoke-passes against current production (all hosts OK, notAfter Aug 23 / Sep 22 2026).

### DIFF
--- a/deploy/k8s/cert-monitor-cronjob.yaml
+++ b/deploy/k8s/cert-monitor-cronjob.yaml
@@ -1,0 +1,162 @@
+##############################################################################
+# Eurobase TLS-cert monitor → Discord #alerts
+#
+# Hourly probe of the live TLS handshake on each documented host. Cert
+# < 14 days → Discord warning. Already expired (or host unreachable) →
+# Discord @everyone. Same exit-code scheme as scripts/ops/cert-status.sh
+# (which this CronJob is the in-cluster sibling of):
+#   0 → quiet success; 1 → WARN post; 2 → CRITICAL post.
+#
+# Why an in-cluster CronJob rather than just cert-manager events:
+#   The June 24 outage (issue: cert silently failed to renew for 30 days
+#   while cert-manager's `Issuing` condition stayed True) showed that
+#   trusting any single piece of the renewal pipeline is wrong. This
+#   check probes the *served* certificate end-to-end — it catches:
+#     * renewal failure (what bit us in June)
+#     * Secret renewed but ingress controller didn't pick it up
+#     * DNS pointed at the wrong cluster (cert is fresh somewhere else)
+#   None of those would be caught by a cert-manager-internal alert.
+#
+# Webhook URL is read from the `eurobase-secrets` Secret key
+# DISCORD_ALERTS_WEBHOOK (optional — if unset, the job logs to its pod
+# logs and exits non-zero so the failed-Job count still moves; adding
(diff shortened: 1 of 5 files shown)


STGime (commented):
## Review — TLS cert monitor + runbook

**Strong incident response.** Probing the **served** certificate via `openssl s_client` (rather than trusting `Certificate.status`) is exactly the right lesson from June 24 — it catches renewal failure, secret-not-picked-up, and wrong-cluster-DNS, none of which an internal-state alert would. The shared exit-code body across `cert-status.sh` (laptop/CI) and the CronJob is clean, and naming `issuerRef` explicitly (no "default" issuer) directly addresses the root cause.

### Worth a conscious sign-off: the Cloudflare sovereignty exception
The substantive prod change is repointing the wildcard `Certificate` from the Scaleway issuer to `letsencrypt-prod-dns-cloudflare` (native Cloudflare DNS-01). **Cloudflare is explicitly on `CLAUDE.md`'s "no US cloud services" list.** Scope is limited — it's ACME challenge automation (TXT writes), and the `eurobase.app` zone is *already* on Cloudflare (that mismatch was the outage's root cause), so no customer data plane is involved — and it's documented as a temporary exception until the apex zone migrates to Scaleway DNS, with the dormant Scaleway issuer kept for a one-line swap. That's the honest handling; I'd just make sure the "migrate apex DNS to Scaleway" step is a tracked issue, not only a runbook paragraph, since this is the same class of deviation as the Discord exception and the product is sovereignty-positioned.

### Minor (non-blocking)
- The CronJob's `alpine:3.20` isn't digest-pinned. Low risk here (unprivileged probe pod, no hostPath — unlike the privileged init in #116 which correctly pins by digest), but pinning would match that precedent.
- A `WARN` (`worst=1`, cert <14d but still valid) exits non-zero, so the Job shows `Failed` on a non-urgent warning. Deliberate visibility choice (mirrors health-monitor) — just flagging it's not a true failure.

Otherwise LGTM (comment, own PR).

---

PR #221, opened 2026-06-25
fix #206: seal edge_functions.env_vars at rest with per-tenant vault key

### LINKED ISSUE #206: edge_functions.env_vars are plaintext at rest, contradicting the documented "encrypted at rest (vault key)" contract
Split out from #201's closing note.

## Discrepancy

`docs/architecture-edge-functions.md` documents the data model as:

```sql
env_vars JSONB,  -- encrypted at rest (vault key)
```

In reality, per-function env vars are plaintext end to end:

- `internal/functions/service.go` stores `UpdateRequest.EnvVars` (`map[string]string`) directly into `edge_functions.env_vars` JSONB — no seal step.
- `functions-runner/server.ts` `loadFunction` reads the raw JSONB and passes it straight into the worker as `ctx.env` — no decrypt step (there is none to do).

So anything a developer puts in env vars — and env vars are the conventional place for API keys — sits unencrypted in `public.edge_functions`, visible to any role/path with SELECT on that table, in DB backups/exports, and in `GET /platform/projects/{id}/functions/{name}` responses.

Meanwhile the vault path is now fully working from functions (#201 fixed per-tenant key decryption in the runner), so the platform has two "give my function a secret" mechanisms with very different at-rest guarantees and no signal which one a developer should use.

## Options

1. **Encrypt env_vars at rest with the per-tenant vault key.** Seal the JSONB (or each value) in `Service.Create/Update` using `internal/vault`'s provider (per-tenant HKDF, `key_version`-tagged — same scheme as `vault_secrets`); decrypt in the runner, which since #201 already has version-aware HKDF decryption (`functions-runner/vault.ts`) that could be reused. Needs a migration (nonce + key_version alongside `env_vars`, or a single sealed blob column) plus a backfill that seals existing rows, and masking in the `GET` function response (the console edit flow needs a decision: re-enter values vs masked echo, cf. the OAuth `client_secret` handling).
2. **Fix the doc + steer to vault.** Document env_vars as plaintext configuration (non-secret toggles, URLs), point secrets at `ctx.vault.get`, and optionally warn at deploy time when an env value looks like a credential (entropy/name heuristics — `*_KEY`, `*_SECRET`, `*_TOKEN`).
3. **Deprecate env_vars for secrets entirely** and inject selected vault secrets as env at invocation time (`vault_refs: ["MISTRAL_API_KEY"]`-style), so there's one sealed store and one DX.

Option 1 honors the doc as written; option 2 is the honest quick fix; option 3 is the cleanest long-term DX but the biggest change. Given #189's precedent (docs promised it → make it true), option 1 or 3 seems right, with 2 as an interim stopgap.

STGime:
Closes #206.

### DIFF
--- a/docs/architecture-edge-functions.md
+++ b/docs/architecture-edge-functions.md
@@ -144,7 +144,20 @@ CREATE TABLE public.edge_functions (
     entrypoint      TEXT NOT NULL DEFAULT 'handler',  -- exported function name
     verify_jwt      BOOLEAN NOT NULL DEFAULT true,    -- require end-user JWT
     import_map      JSONB,                  -- optional Deno import map
-    env_vars        JSONB,                  -- encrypted at rest (vault key)
+    -- env vars at rest (#206):
+    --   New writes seal the JSON({K:V}) blob with the per-tenant
+    --   HKDF-derived AES-256-GCM key (same KDF as vault_secrets, see
+    --   internal/vault/keyprovider.go). The trio below is all-or-nothing
+    --   per row (CHECK constraint). Legacy `env_vars JSONB` is plaintext
+    --   and is lazy-migrated: any Update of a row seals the values into
+    --   env_vars_blob and NULLs the legacy column. Once the legacy
+    --   column is NULL everywhere it can be dropped in a follow-up
+    --   migration. The runner prefers sealed; falls back to legacy
+    --   while it's still populated.
+    env_vars              JSONB,            -- DEPRECATED: legacy plaintext (lazy-migrated)
+    env_vars_blob         BYTEA,            -- AES-256-GCM ciphertext of JSON({K:V})
+    env_vars_nonce        BYTEA,            -- AEAD nonce (12 bytes)
+    env_vars_key_version  SMALLINT,         -- key version used to seal (HKDF info: "eurobase-vault-v<n>")
     status          TEXT NOT NULL DEFAULT 'active' CHECK (status IN ('active','disabled')),
     version         INTEGER NOT NULL DEFAULT 1,
     created_at      TIMESTAMPTZ DEFAULT now(),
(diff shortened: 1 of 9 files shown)


STGime (commented):
## Review — seal `edge_functions.env_vars` at rest

**Solid, and secure by construction.** Verified the design closes the plaintext-at-rest gap without a backfill footgun. `go build` + `internal/functions` tests pass; migration `000067` is sequential (main is at `000066`).

### Why the contract holds
- **No plaintext write path.** `CreateRequest` has no `env_vars` field and `Create`'s INSERT doesn't touch the column, so env vars can only be set via `Update` — which always goes through `sealEnvVars` (per-tenant HKDF → AES-256-GCM, reusing the proven `vault` pipeline via the new `SealForTenant`/`OpenForTenant`). The only plaintext path is the explicit vault-disabled (dev) fallback, which logs a loud per-write warning.
- **No re-leak.** `Update` clears the legacy `env_vars` column whenever it seals, so a sealed row never keeps plaintext.
- **Partial-write safe.** The `edge_functions_env_vars_sealed_consistent` CHECK (all three sealed columns NULL together or NOT NULL together) plus `nullIfZero(keyVersion)` guarantees no undecryptable row.
- **Fail-safe reads.** Both Go and the Deno runner prefer sealed and never fall back to legacy plaintext on a decrypt failure — Go surfaces an error, the runner logs and runs with an empty env. Good call: a key mismatch is visible, not silently papered over with the deprecated column.

### Minor (non-blocking)
- `schemaNameFor`'s comment says it's "cached briefly per (projectID)" but it does a fresh `SELECT schema_name` on every call — either add the cache or drop the comment (it's an extra query per sealed Get/GetByID; the runner side avoids it via its JOIN).
- Go `resolveEnvVars` returns an error on decrypt failure while the runner returns `{}` — both avoid the legacy fallback (intended), just noting the divergence.
- The runner caches the resolved function incl. env, so a decrypt-failure empty-env can persist in cache until eviction after a key is healed — worth a note for the rotation runbook.

No blockers — LGTM (comment, own PR).

---

Issue #224, opened 2026-06-26
[v1.3] Per-project Rate Limits (Auth + Email/SMS) — Supabase-parity (umbrella)


STGime:
**Umbrella for the per-project Rate Limits page.** Goal: let each project's owner safeguard their own quota knobs from a console page, mirroring Supabase's published Rate Limits UI. Sensible defaults out of the box; per-project overrides via the existing `auth_config` JSONB.

## Background
Today every rate limit on the auth surface is a **global hardcoded constant** (`internal/ratelimit/auth.go`); email and SMS sends have **zero platform-level quotas** — only the Scaleway TEM and GatewayAPI provider limits stand between a misconfigured tenant and a bill. Same risk class as the 7% bounce rate caught 2026-06-25.

## Scope locked
**In**
- emails/h (per project)
- SMS/h (per project)
- token refresh /5min/IP
- token verification /5min/IP (covers OTP + magic-link verify)
- sign-up + sign-in /5min/IP
- "IP Address Forwarding" trust-proxy toggle

**Out**
- Anonymous sign-ups — Eurobase has no anonymous-user concept (anon API key is different)
- Web3 sign-ups — not shipped
- Per-project override of `plan_limits.rate_limit_rps` — data-plane RPS stays plan-tiered

## Architecture in one paragraph
Storage: extend `projects.auth_config JSONB` with a `rate_limits` sub-object — same per-project pattern as `cors_origins`, `password_min_length`, etc. (`internal/tenant/auth_config.go:52`). Defaults as Go constants — absent sub-object means "use defaults"; no backfill. Enforcement: reuse `RateLimiter.Allow(key, limit, window)` (`internal/ratelimit/limiter.go`) — only the **source** of `(limit, window)` changes (constants → per-request lookup). Email/SMS get a separate per-project hourly Redis counter; SMS needs `projectID` threaded through `Service.Send`. IP forwarding: per-project flag controls whether the limiter keys off the leftmost `X-Forwarded-For` entry or the TCP peer.

## Sub-issues
- [ ] **#225** Schema + defaults + refactor existing hardcoded auth limits to per-project (foundation)
- [ ] **#226** New auth limits: token refresh + token verification per-IP gates
- [ ] **#227** Email/SMS per-project hourly quotas + SMS signature refactor for `projectID`
- [ ] **#228** IP Address Forwarding (trust-proxy) toggle + middleware switch
[...]

Issue #228, opened 2026-06-26
[v1.3][rate-limits #D] IP Address Forwarding (trust-proxy) toggle


STGime:
Tracked under #224. Depends on #225.

## What lands
Per-project knob `auth_config.rate_limits.trust_proxy` (default `false`) controlling how the auth rate limiter sources the client IP:

- `trust_proxy = false` (default): use the TCP peer. Safer — `X-Forwarded-For` can be forged when the request hasn't traversed a controlled hop.
- `trust_proxy = true`: use the leftmost `X-Forwarded-For` entry. Tenants who sit behind their own edge (CDN, WAF) opt in so the limiter sees the real client and not the CDN IP.

The platform default ("trust the nginx-ingress hop only") is unchanged for the existing data-API limiter — this knob only affects the **per-project auth** keying, so flipping it in one project doesn't widen attack surface anywhere else.

## Why default off
A new project that goes live without an edge in front of it would be **harmed** by trusting forwarded headers (one attacker IP can rotate XFF and bypass the rate limit). Off-by-default is the safe default; the docs page in #F calls this out explicitly.

## Acceptance
- With `trust_proxy=false`, send 31 `/signup` requests with the same TCP peer but rotating `X-Forwarded-For` → 31st returns 429 (same counter).
- Flip `trust_proxy=true`, repeat → each new XFF starts a fresh counter (limit per forwarded IP, not per TCP peer).
- Existing prod behavior unchanged because every prod project starts with `trust_proxy=false`.

## Files
- `internal/ratelimit/auth.go` — `identifierForIP(r, projectCfg)` helper
- `internal/gateway/router.go` — wiring (the subdomain middleware already resolves the per-project ctx)

🤖 Filed with [Claude Code](https://claude.com/claude-code)

Issue #227, opened 2026-06-26
[v1.3][rate-limits #C] Email + SMS per-project hourly quotas (+ SMS signature refactor)


STGime:
> ## Status (2026-06-26) — split into two shippable halves
>
> The email half hit a structural blocker on review of #234: defaulting `EmailsPerHour: 2` silently breaks confirmation flows for any project with >2 signups/hour, and the platform's single-shared-SMTP model has no escape hatch. **Email enforcement is deferred behind #235** (BYO-SMTP + paid Scaleway-hosted email add-on).
>
> The SMS half is shippable today (projects bring their own GatewayAPI budget via configuration, so a per-project cap is a real safeguard, not a self-DoS). Lives in **PR #234**:
> - `internal/ratelimit/quota.go` helper (supports both `email` and `sms` actions; email path dormant until #235)
> - `AuthService.checkSMSQuota` enforces `auth_config.rate_limits.sms_per_hour` on the phone-OTP path
> - `sms.Service.SendOTP(ctx, projectID, …)` signature change
> - `AuthService.SetRateLimiter` wired in the router
> - Migration 000069 (comment-only; documents the `quota:sms:project:{id}` keyspace)
>
> **Don't close this issue when #234 merges** — it stays open as the home for the email-quota work, which lands when #235 unblocks it. At that point this issue's "Email side" section below becomes the spec for the follow-up PR.
>
> Related: `SignupSigninPer5MinPerIP` default stays at the interim 8 (~96/h) until email enforcement lands; bumps back to 30 (Supabase parity) then.

---

Tracked under #224. Depends on #225.

## What lands
Per-project hourly send quotas on the email and SMS pipelines. The same risk class as the 2026-06-25 TEM bounce-rate finding: a misconfigured or abused tenant today can drain the platform-wide TEM quota and tank everyone else's sender reputation. This adds a hard cap, surfaced as 429 with `Retry-After`.

### Email side
- Wrap `SendVerificationEmail`, `SendPasswordResetEmail`, `SendMagicLinkEmail`, and `SendRaw` in `internal/email/service.go` with a per-project hourly counter check.
- Counter key: `email:project:{id}:hour:{epoch_hour}`. Redis INCR + EXPIRE (existing `RateLimiter` Lua script pattern is reusable).
- Over-quota → return a new typed `ErrEmailQuota`; the handler maps it to 429 with `Retry-After` set to the seconds remaining in the hour.
[...]

PR #231, opened 2026-06-26
fix #225: per-project auth rate-limit foundation

### LINKED ISSUE #225: [v1.3][rate-limits #A] Schema + defaults + refactor existing hardcoded auth limits to per-project
Foundation for the per-project Rate Limits series. Tracked under #224.

## What lands
- New migration `migrations/000068_auth_config_rate_limits.{up,down}.sql` — comment-only, documenting the `auth_config.rate_limits` JSONB sub-object contract (no DDL — schema is already `JSONB`).
- Extend `AuthConfig` (`internal/tenant/auth_config.go:52`) with a `RateLimits` sub-struct (omitempty) + a `DefaultRateLimits` exported value:
  - `signup_signin_per_5min_per_ip` = 30 (mirrors current 5/h aggregate ≈ 30 within 5-min burst window)
  - `emails_per_hour` = 2 (Supabase default — intentionally tight)
  - `sms_per_hour` = 30 (Supabase default)
  - `token_refresh_per_5min_per_ip` = 150 (covered in #B)
  - `token_verification_per_5min_per_ip` = 30 (covered in #B)
  - `trust_proxy` = false (covered in #D)
- Drop the hardcoded constants in `internal/ratelimit/auth.go`. Introduce `CheckAuthRateForProject(ctx, action, projectCtx, identifier)` that:
  1. Resolves `(limit, window)` from the project's `auth_config.rate_limits.<knob>`, falling back to `DefaultRateLimits` when nil/zero.
  2. Composes the Redis key as `auth:{action}:project:{projectID}:{identifier}` so two tenants never collide.
  3. Delegates to the existing `RateLimiter.Allow` — fail-open behavior preserved when Redis is unreachable.
- Swap the six call sites in `internal/enduser/handler.go` (signup, signin, forgot-password, magic-link, resend-verify, phone OTP) from `CheckAuthRate` (constants) to `CheckAuthRateForProject`.

## Acceptance
- `go test ./internal/ratelimit/... -count=1` green.
- New project's `auth_config` row has no `rate_limits` sub-object — defaults apply via fallback.
- Hit `/v1/auth/signup` past the default limit from a single IP → 429 (current behavior preserved).
- `PATCH /v1/tenants/{id}` setting `auth_config.rate_limits.signup_signin_per_5min_per_ip = 1` makes the next attempt 429 immediately.

## Files
- `migrations/000068_auth_config_rate_limits.up.sql` (+ `.down.sql`)
- `internal/tenant/auth_config.go`
- `internal/ratelimit/auth.go`
- `internal/enduser/handler.go` (6 call sites)

STGime:
Foundation for #224 series. Migration 000068 + `auth_config.rate_limits` + `CheckAuthRateForProject` + signup/signin per-IP gate. Signup default loosens from 5/h to 30/5min/IP (Supabase parity); per-email brute-force counter unchanged. Tests for merge semantics + tenant isolation. Closes #225.

### DIFF
--- a/internal/enduser/handler.go
+++ b/internal/enduser/handler.go
@@ -32,13 +32,17 @@ func HandleSignUp(svc *AuthService, limiter ...*ratelimit.RateLimiter) http.Hand
 
 		config := tenant.ParseAuthConfig(pc.AuthConfig)
 
-		// Per-project per-IP volume gate covering signup. Default 30
-		// requests / 5 min / IP — overridable via the Rate Limits page
-		// (#229). The per-account anti-brute-force gates (e.g. the
-		// signin-failure counter keyed by email, the platform-wide
+		// Per-project per-IP volume gate covering signup. Same knob as
+		// the signin handler below — overridable via the Rate Limits
+		// page (#229). The per-account anti-brute-force gates (e.g.
+		// the signin-failure counter keyed by email, the platform-wide
 		// resend-verify rate limit) are a separate axis and stay at
 		// platform defaults — a project loosening this knob does not
 		// loosen those.
+		//
+		// XFF caveat: see the signin handler below — until #228 lands
+		// the `trust_proxy` enforcement, this gate trusts the
+		// leftmost X-Forwarded-For unconditionally.
 		rlCfg := config.EffectiveRateLimits()
 		if ratelimit.CheckAuthRateForProject(rl, w, r.Context(), "signup_signin", pc.ProjectID, ratelimit.ClientIP(r), rlCfg.SignupSigninPer5MinPerIP, ratelimit.FiveMinutes) {
 			return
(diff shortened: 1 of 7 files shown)


STGime (commented):
## Review — per-project auth rate-limit foundation

**Clean foundation, well-tested; one security-relevant default to dial back before merge.** The config/merge layer and per-project keying are solid — the thing to weigh is that this loosens the signup volume gate substantially *ahead* of the compensating controls in the same series.

### Strengths
- **Merge semantics right + well-tested** — `EffectiveRateLimits` treats `0` as "use default" (avoids the `emails_per_hour: 0` self-DoS footgun), `TrustProxy` is set-only (never promoted to a default), and there's a JSON round-trip test guarding the tags.
- **Per-project keying is correct** — `auth:{action}:project:{projectID}:{identifier}` is namespaced so two tenants on a shared NAT'd IP get independent counters (explicitly tested) and won't alias the legacy key shape during rollout.
- **Right call keeping anti-brute-force gates separate** — the per-email signin-failure counter and the platform identifier-keyed gates stay at platform defaults as security floors, not knobs. That distinction is reasoned correctly.
- Comment-only migration for a JSONB shape extension is a sensible no-backfill approach.

### Recommend a lower default: ~100/hour/IP (not 360/hour)
The default goes from **5/hour/IP → 30/5min/IP (= 360/hour/IP)**, a 72× loosening. The justification (Supabase parity, per-email brute-force counter unchanged) holds *in the end state*, but the compensating controls that make 360/h safe — `EmailsPerHour` (default 2) / `SMSPerHour` — are defined in this struct but explicitly **"Wired in #227", not enforced here**. So in the window between this PR and #227, one IP can drive ~360 signups/hour, each firing a verification email (vs 5 before), with no email cap behind it — a spam / email-cost / sender-reputation risk.

5/hour was too tight (a shared office IP trips it immediately), but 360/hour is too loose to ship before the email cap. **Suggest setting the default to ~100/hour/IP** — concretely, since the knob is per-5-min, `SignupSigninPer5MinPerIP: 8` (≈96/h) — as a middle ground: ~20× more usable than today, but ~3.6× tighter blast radius than 360/h until #227 lands. Then bump it back to 30/5min in the #227 PR once the email/SMS caps actually enforce. (The per-email signin-fail counter still protects *signin* brute-force; this is specifically the *signup-volume / email-amplification* axis.)

### Secondary (pre-existing, amplified by the loosening)
- **The per-IP key trusts forged `X-Forwarded-For`.** `ratelimit.ClientIP` returns the leftmost XFF unconditionally; the new `TrustProxy` knob (the intended fix, "#228") isn't consulted yet. So an attacker rotating XFF gets a fresh counter per request unless the nginx ingress *overwrites* (not appends) XFF. Pre-existing keying, but a looser-and-still-bypassable gate is weaker than what it replaced — worth confirming the ingress overwrites XFF, and landing #228 close behind.

### Nit
- `rl_signin` in `HandleSignIn` is snake_case (non-idiomatic Go, inconsistent with `rlCfg` in the sibling `HandleSignUp`).

Net: foundation is merge-quality; I'd just lower the signup default to ~100/h/IP until #227's email cap lands.

---

PR #232, opened 2026-06-26
fix #226: token refresh + verification per-IP gates

### LINKED ISSUE #226: [v1.3][rate-limits #B] Token refresh + token verification per-IP gates

## What lands
Two new auth rate-limit knobs, both per-IP per-5-minutes, plumbed through the same `CheckAuthRateForProject` path landed in #225.

- **Token refresh** (`/v1/auth/refresh`). Default 150/5min/IP — generous because legitimate SDK clients refresh proactively; mirrors Supabase. Inside the existing refresh handler, gate on the IP (with the per-project trust-proxy switch landing in #D).
- **Token verification** (covers `/v1/auth/verify-otp` and `/v1/auth/verify-magic-link`). Default 30/5min/IP. Throttles OTP brute-force.

The knobs land as fields on `auth_config.rate_limits` (already added in #225); per-project override via the same console flow.

## Why this matters separately
- Refresh-token endpoint today is **unrate-limited**. A bad actor with one valid refresh token can mint access tokens until the family is revoked (#56 is the longer-term fix; this is the burst guard).
- OTP verify today is **unrate-limited at the verify step** — only the OTP issue step is gated. Without a verify limit, a 6-digit OTP is brute-forceable within 5 minutes.

## Acceptance
- POST `/v1/auth/refresh` 151 times in 5 min from one IP → 151st returns 429.
- POST `/v1/auth/verify-otp` 31 times in 5 min → 31st returns 429.
- Per-project override via `auth_config.rate_limits.token_refresh_per_5min_per_ip` takes effect on the next request.
- Existing token rotation behavior unchanged below the limit.

## Files
- `internal/enduser/handler.go` — refresh handler + new verify gate
- `internal/ratelimit/auth.go` — two new action constants if not already added in #225

STGime:
Closes #226 (umbrella #224). Wires the TokenRefresh and TokenVerification knobs landed in #225 into HandleRefresh, HandleVerifyEmail, HandleSignInWithMagicLink, HandleVerifyPhoneOTP. Same per-project keying + XFF caveat (tracked in #228). See commit body for design.

### DIFF
--- a/internal/enduser/handler.go
+++ b/internal/enduser/handler.go
@@ -138,7 +138,11 @@ func HandleSignIn(svc *AuthService, limiter ...*ratelimit.RateLimiter) http.Hand
 }
 
 // HandleRefresh returns an HTTP handler for POST /v1/auth/refresh.
-func HandleRefresh(svc *AuthService) http.HandlerFunc {
+func HandleRefresh(svc *AuthService, limiter ...*ratelimit.RateLimiter) http.HandlerFunc {
+	var rl *ratelimit.RateLimiter
+	if len(limiter) > 0 {
+		rl = limiter[0]
+	}
 	return func(w http.ResponseWriter, r *http.Request) {
 		pc, ok := auth.ProjectFromContext(r.Context())
 		if !ok {
@@ -147,6 +151,22 @@ func HandleRefresh(svc *AuthService) http.HandlerFunc {
 		}
 
 		config := tenant.ParseAuthConfig(pc.AuthConfig)
+
+		// Per-project per-IP gate on the refresh endpoint (#226 —
+		// closes the burst-guard gap that #56's reuse-detection
+		// can't cover before a stolen token is family-revoked).
+		// Default 150 per 5 min — generous because legitimate SDK
+		// clients refresh proactively; tighten via the Rate Limits
+		// page.
+		//
+		// XFF caveat: ratelimit.ClientIP trusts the leftmost
+		// X-Forwarded-For unconditionally; the per-project
+		// trust_proxy knob enforcement lands in #228. Prod traffic
+		// relies on the nginx-ingress XFF rewrite in the meantime.
+		rlCfg := config.EffectiveRateLimits()
+		if ratelimit.CheckAuthRateForProject(rl, w, r.Context(), "token_refresh", pc.ProjectID, ratelimit.ClientIP(r), rlCfg.TokenRefreshPer5MinPerIP, ratelimit.FiveMinutes) {
+			return
+		}
 
 		var req RefreshRequest
 		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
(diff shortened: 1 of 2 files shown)


STGime (commented):
## Review — token refresh + verification per-IP gates

**The wiring is clean and consistent (fine to merge as-is), but the headline claim — "closes the OTP/magic-link brute-force gap" — is overstated. A per-IP gate can't close it, and the real phone-OTP weakness needs a separate, stronger fix.**

### The mechanical change is good
Wires the `TokenRefresh`/`TokenVerification` knobs from #225 into `HandleRefresh`, `HandleVerifyEmail`, `HandleSignInWithMagicLink`, `HandleVerifyPhoneOTP` with the same `CheckAuthRateForProject` keying, the same variadic-`limiter` handler pattern, and the same documented XFF caveat; the router threads `limiter` to all four. Consistent with the series; build + `enduser`/`ratelimit` tests pass. As incremental defense-in-depth, ✅.

### Main finding: the phone-OTP gap isn't closed by this
For phone OTP specifically, a per-IP rate limit isn't an adequate control, because `internal/sms/service.go` `VerifyOTP` is weak in two pre-existing ways:

1. **The code isn't bound to the phone/user.** Verify is `UPDATE … WHERE token_hash = sha256(code) AND token_type='phone_verification' AND expires_at > now() AND used_at IS NULL RETURNING user_id` — **tenant-wide, code-only**. A guessed 6-digit code matches *any* outstanding OTP in the tenant and returns that user's id → login as an arbitrary user. With K OTPs outstanding, expected guesses to hit *someone* ≈ 10⁶/K, well under the 10⁶ a targeted attack would need.
2. **No per-token / per-phone attempt counter** — the only bound is single-use + 5-min expiry.

Against that, the only new control here is **per-IP, default 30/5min, and the IP is forgeable** (the documented `ClientIP`→leftmost-`X-Forwarded-For`; `trust_proxy`/#228 not wired yet). An attacker rotating `X-Forwarded-For` gets unlimited attempts; even with honest IPs, a small botnet clears 30/5min against a code-only match. So the gap is **narrowed, not closed**.

**Real fix (separate from this PR's scope):** bind OTP verify to the target phone (`WHERE phone = $1 AND token_hash = $2 …`) **and** add a per-token attempt counter that invalidates the OTP after ~5 failures (optionally widen the code). That's what actually bounds a 6-digit space; the per-IP gate is defense-in-depth on top. Suggest softening the commit/comment wording so it doesn't read as "closed" (false assurance), and tracking the `VerifyOTP` fix as its own security issue.

(Email-verify, magic-link, and reset tokens are high-entropy hex via `rand.Read`, so for those the rate limit really is just defense-in-depth — phone OTP is the one with a real residual.)

### Minor
- **`reset-password` is left ungated** — it consumes a `password_reset` token like the other verify endpoints, so for consistency it should share the `token_verify` gate. Low risk (high-entropy token, not enumerable), just an inconsistency.
- The three verify endpoints share one `token_verify` per-IP bucket (documented, fine) — mixed verify traffic from one IP draws down a shared budget.

Net: mergeable for the incremental per-IP coverage, but don't treat phone-OTP brute-force as resolved — the code-only, attempt-unlimited `VerifyOTP` is the actual vulnerability and deserves its own fix.

---

PR #234, opened 2026-06-26
fix #227 (partial): per-project SMS hourly quota + projectID signature

### LINKED ISSUE #227: [v1.3][rate-limits #C] Email + SMS per-project hourly quotas (+ SMS signature refactor)

STGime:
Closes the SMS half of #227 (umbrella #224). The email half is deferred behind the BYO-SMTP / paid-TEM unblocker tracked in **#235**.

## What lands
- **Quota helper** (`internal/ratelimit/quota.go`) — `CheckProjectHourlyQuota` + `ErrQuotaExceeded` sentinel. Per-project + per-action keyspace, fail-open on nil limiter / Redis error, atomic INCR via the existing Lua script.
- **SMS quota enforcement** — `AuthService.checkSMSQuota` reads `auth_config.rate_limits.sms_per_hour` (default 30) and silently skips over-quota sends with a `slog.Warn` operator signal. Phone-OTP path returns 200 like the existing unknown-phone case so a 429 can't be used to probe whether a phone is in the system.
- **`sms.Service.SendOTP(ctx, projectID, schemaName, userID, phone)`** — `projectID` threaded through per the umbrella issue. Used today as a metering anchor; future work can move the quota check down without another signature break.
- **`AuthService.SetRateLimiter`** wired in the router so the limiter reaches the AuthService.
- **Migration 000069 (comment-only)** documents the Redis key shape (`quota:sms:project:{projectID}`).
- **Tests** — `quota_test.go` covers nil-fails-open, per-project isolation, and email-vs-SMS action keyspace independence (the helper supports both even though only SMS is wired today).

## What was stripped after the review
- Email enforcement on `SignUp` verification, `ForgotPassword`, `RequestMagicLink`, `ResendVerification` — the platform's single-shared-SMTP model means a default `EmailsPerHour: 2` silently breaks confirmation flows for any project with >2 signups/hour, and there's no BYO-SMTP escape hatch. Tracked in **#235** (per-project BYO-SMTP + paid Scaleway-hosted email add-on).
- `SignupSigninPer5MinPerIP` stays at the interim **8** (≈96/h) from #231 — the per-IP loosening to 30 requires the email cap to be live to bound the amplification path. Bump returns once #235 lands.
- `EmailsPerHour` and `checkEmailQuota` references in `AuthService` — the knob still exists in `AuthConfig` so the UI can surface it once enforcement returns, but no code path consumes it today.

## Behaviour notes
- Quota-exceeded paths **return their normal success status** with a `slog.Warn` as the operator signal. No 429 on send-side caps — those would leak email/phone enumeration.
- Migration doc key shape now matches the actual code (was `quota:{action}:project:{projectID}:hour:{epoch_hour}` in the doc; code uses `quota:{action}:project:{projectID}` — the hour bucketing is Redis-TTL-driven, not in the key).

Tests: `go test ./internal/{tenant,ratelimit,enduser,gateway,email,sms}/... -count=1` green.

### DIFF
--- a/internal/enduser/auth_service.go
+++ b/internal/enduser/auth_service.go
@@ -15,6 +15,7 @@ import (
 	"github.com/eurobase/euroback/internal/db"
 	"github.com/eurobase/euroback/internal/email"
 	"github.com/eurobase/euroback/internal/oauth"
+	"github.com/eurobase/euroback/internal/ratelimit"
 	"github.com/eurobase/euroback/internal/sms"
 	"github.com/eurobase/euroback/internal/tenant"
 	"github.com/golang-jwt/jwt/v5"
@@ -56,6 +57,12 @@ type AuthService struct {
 	emailService *email.EmailService
 	smsService   *sms.Service
 	oauthSecrets OAuthSecretLookup
+
+	// rateLimiter (#227) gates per-project hourly send quotas on email
+	// and SMS paths. Nil-safe: a missing limiter means quota checks
+	// fail open (same fail-open contract as the auth limiter — a Redis
+	// outage shouldn't take auth offline). Set via SetRateLimiter.
+	rateLimiter *ratelimit.RateLimiter
 }
 
 // NewAuthService creates a new end-user auth service.
@@ -71,6 +78,34 @@ func (s *AuthService) SetEmailService(svc *email.EmailService) {
 // SetSMSService sets the SMS service for sending phone OTP codes.
 func (s *AuthService) SetSMSService(svc *sms.Service) {
 	s.smsService = svc
+}
+
+// SetRateLimiter wires in the Redis-backed limiter used for per-project
+// hourly quotas on email and SMS sends (#227). Without it, every
+// `checkEmailQuota` / `checkSMSQuota` call fails open — the contract
+// matches the auth limiter so dev environments without Redis stay
+// functional.
+func (s *AuthService) SetRateLimiter(rl *ratelimit.RateLimiter) {
+	s.rateLimiter = rl
+}
+
+// checkSMSQuota gates a per-project SMS send on the hourly cap from
+// auth_config.rate_limits.sms_per_hour. Returns nil if the send may
+// proceed; returns ratelimit.ErrQuotaExceeded when the project is over
+// its hourly cap. Same fail-open contract as the auth limiter (nil
+// limiter / Redis hiccup → allow).
+//
+// Email side is NOT yet enforced — per the #234 review, the platform's
+// single-shared-SMTP model means an aggressive per-project email cap
+// breaks legitimate signup flows with no BYO-SMTP escape hatch. The
+// EmailsPerHour knob still exists in the config struct so the UI can
+// surface it once BYO-SMTP lands; until then the quota helper is wired
+// only for SMS, where projects bring their own per-tenant GatewayAPI
+// budget via configuration.
+func (s *AuthService) checkSMSQuota(ctx context.Context, projectID string, cfg tenant.AuthConfig) error {
+	limits := cfg.EffectiveRateLimits()
+	_, err := ratelimit.CheckProjectHourlyQuota(s.rateLimiter, ctx, "sms", projectID, limits.SMSPerHour)
+	return err
 }
 
 // SetOAuthSecretLookup wires in a function that resolves encrypted OAuth
@@ -843,7 +878,7 @@ func (s *AuthService) ResendVerification(ctx context.Context, schemaName, projec
 }
 
 // SendPhoneOTP finds or creates a user by phone number and sends an OTP code.
-func (s *AuthService) SendPhoneOTP(ctx context.Context, schemaName, phone string) error {
+func (s *AuthService) SendPhoneOTP(ctx context.Context, schemaName, projectID, phone string, config tenant.AuthConfig) error {
 	if s.smsService == nil {
 		return fmt.Errorf("sms not available: SMS provider not configured")
 	}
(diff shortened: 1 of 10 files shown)


STGime (commented):
## Review — email + SMS per-project hourly quotas

**The quota plumbing is clean and well-tested, but the `EmailsPerHour: 2` default is a blocker — it silently breaks account-confirmation email for any real project, and there's no escape hatch.**

### Strengths
- `CheckProjectHourlyQuota` is solid: per-project + per-action key isolation (both tested), fail-open on a nil limiter *and* on Redis error (an outage can't take auth offline), atomic INCR with a clean `ErrQuotaExceeded` sentinel.
- Enumeration protection preserved — forgot-password / magic-link / phone-OTP silently skip + `slog.Warn` + still return 200.
- Signup revert 8→30 is internally consistent and well-documented (matches the #231 plan); migration `000069` is sequential + comment-only; good tests.

### Blocker: default of 2 auth emails/hour/project, shared across all email types, silently dropped
Traced end to end:
- `DefaultRateLimits().EmailsPerHour = 2`.
- `checkEmailQuota` → `CheckProjectHourlyQuota(rl, ctx, "email", projectID, 2)`; the Redis key is `quota:email:project:{projectID}` — **project-wide, no recipient dimension**.
- `SignUp` (verification), `ForgotPassword`, `RequestMagicLink`, `ResendVerification` all share that one `"email"` counter.
- `limit=2` ⇒ 2 allowed/hour, 3rd blocked; over-quota is a silent skip (`slog.Warn`, handler still 200).

So **by default a project can send only 2 confirmation emails per hour — and that budget is shared with password resets and magic links.** The 3rd user to sign up in an hour is created but never receives a verification email and can't complete signup. No user-facing signal.

This is "Supabase parity" on the number only. Supabase's 2/hour is the **built-in-SMTP** limit and you lift it by configuring your own SMTP. Eurobase has **no BYO-SMTP path** (verified — everything goes through the single platform `EmailService`), so 2/hour is a hard floor whose only escape is editing the `emails_per_hour` override. As a default, that's not shippable.

**Required before merge:** raise the email default to a realistic platform-sender value, and split **verification** onto its own (higher) bucket so a burst of resets/magic-links can't starve account confirmations — or don't enforce the email cap until BYO-SMTP exists. Opening a separate feature issue for per-project custom SMTP + a paid Scaleway-hosted email add-on (the proper way for projects to raise their own ceiling).

### Minor
- Quota burns on *attempt*, not success — a transient TEM failure still consumes a slot; painful at a 2/hour budget.
- `sms.Service.SendOTP` gains an unused `projectID` (reserved, documented) — fine.
- Migration `000069` doc text says the key is `quota:{action}:project:{projectID}:hour:{epoch_hour}`, but the code key is `quota:{action}:project:{projectID}` (window passed to `limiter.Allow`). Harmless, but align the doc.

Net: per-project quota mechanics are merge-quality; the 2/hour email default + silent-drop must change before this lands.

---

Issue #233, opened 2026-06-26
[security] Phone OTP verify matches code tenant-wide, no per-token attempt counter


STGime:
**Surfaced during #232 review.** The phone-OTP verify path is weaker than a 6-digit code space suggests, in two compounding ways.

## Bug

`internal/sms/service.go` `VerifyOTP` runs roughly:

```sql
UPDATE phone_otps
SET used_at = now()
WHERE token_hash = sha256(code)
  AND token_type = 'phone_verification'
  AND expires_at > now()
  AND used_at IS NULL
RETURNING user_id
```

Two problems:

1. **No phone binding.** A guessed 6-digit code matches *any* outstanding OTP in the tenant — the returned `user_id` is whichever account that hash belongs to. With K OTPs outstanding at any moment, expected guesses to land *on someone* ≈ 10⁶/K, well below the 10⁶ a targeted attack would need.
2. **No per-token attempt counter.** The only bound on guesses is the 5-min TTL + single-use (`used_at IS NULL`). So an attacker has 5 minutes × however many concurrent attempts they can run.

The per-IP rate limit in #232 (default 30 / 5 min / IP, shared `token_verify` bucket) **narrows** the window — but XFF is forgeable until #228 lands, and a modest botnet still clears 30/5min easily.

## Proper fix (separate from #232)

1. **Bind verify to the target phone:** `WHERE phone = $1 AND token_hash = $2 …`. Phone is already supplied in the verify payload; no API change needed, just the WHERE clause.
2. **Per-token attempt counter:** add an `attempts INT NOT NULL DEFAULT 0` column on `phone_otps`. Each verify increments it; when `attempts >= 5` invalidate the OTP (set `used_at`). Combined with the 5-min TTL and single-use, this bounds an attacker to 5 guesses per outstanding OTP regardless of IP.
3. **Optional:** widen the code (e.g. 8 digits) — but #1 + #2 above are the load-bearing fixes; the code width is secondary.

Email-verify, magic-link, and password-reset tokens are unaffected — they're high-entropy hex from `crypto/rand`, not enumerable.

## Severity

- **P1-should-have** rather than P0 because exploitation requires (a) at least one outstanding phone-OTP in the tenant, (b) a botnet or forged XFF for any meaningful guess rate. Real but bounded by the small phone-auth surface today.
- Worth fixing before phone auth sees broader adoption.

## Related

- #232 — the per-IP defense-in-depth (this issue tracks the underlying SQL/attempt-counter fix).
[...]

PR #237, opened 2026-06-27
fix #228: per-project trust_proxy enforcement (default true for Eurobase ingress)

### LINKED ISSUE #228: [v1.3][rate-limits #D] IP Address Forwarding (trust-proxy) toggle

STGime:
Closes #228 (umbrella #224). Per-project IP-source switch + flips the default to true (Supabase divergence) because Eurobase nginx-ingress rewrites XFF — false would collapse every projects per-IP counter to per-project-total. Detailed deployment rationale lives in three doc blocks (RateLimits.TrustProxy field, DefaultRateLimits, ClientIPForProject helper). See commit body for the full story.

### DIFF
--- a/internal/enduser/handler.go
+++ b/internal/enduser/handler.go
@@ -40,11 +40,12 @@ func HandleSignUp(svc *AuthService, limiter ...*ratelimit.RateLimiter) http.Hand
 		// platform defaults — a project loosening this knob does not
 		// loosen those.
 		//
-		// XFF caveat: see the signin handler below — until #228 lands
-		// the `trust_proxy` enforcement, this gate trusts the
-		// leftmost X-Forwarded-For unconditionally.
+		// The IP source honours auth_config.rate_limits.trust_proxy
+		// (#228): TCP peer by default, leftmost X-Forwarded-For when
+		// the project explicitly opts in. See ClientIPForProject for
+		// the trade-off.
 		rlCfg := config.EffectiveRateLimits()
-		if ratelimit.CheckAuthRateForProject(rl, w, r.Context(), "signup_signin", pc.ProjectID, ratelimit.ClientIP(r), rlCfg.SignupSigninPer5MinPerIP, ratelimit.FiveMinutes) {
+		if ratelimit.CheckAuthRateForProject(rl, w, r.Context(), "signup_signin", pc.ProjectID, ratelimit.ClientIPForProject(r, rlCfg.TrustProxy), rlCfg.SignupSigninPer5MinPerIP, ratelimit.FiveMinutes) {
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


STGime (commented):
## Review — per-project `trust_proxy` enforcement (default true)

**The code is correct and cleanly done; the risk is the *default-true* decision, which rests on an ingress assumption this repo doesn't verify. Worth confirming before relying on it, because if the assumption is wrong the default makes every per-IP gate either ineffective or bypassable.**

### Strengths
- **`bool → *bool` is the right fix** — distinguishes "explicit false" from "field absent," `EffectiveRateLimits` collapses nil→default so `*rlCfg.TrustProxy` is panic-safe (verified every return path sets it), and the migration is JSON-clean: the old `omitempty bool` only ever serialized `true` (false was omitted), so no existing project is stuck on a stale `false`.
- **`ClientIPForProject` is correct and well-tested** — ignores XFF entirely when `false` (the anti-forgery side), leftmost-XFF when `true`, TCP-peer fallback, single-entry case; legacy `ClientIP` preserved for the audit-log path with a snapshot test.
- The "`false` collapses every user onto the nginx pod IP" reasoning is real and correct; comments are thorough.

### Main finding: default-true is safe only under an unverified infra assumption
Keying on **leftmost `X-Forwarded-For`** is correct only if exactly one trusted hop authoritatively overwrites XFF with the real client IP. The PR argues this for "bare nginx-ingress with `use-forwarded-headers: false`," but the deployment has a Scaleway LB in front of nginx, and what actually lands in XFF depends on the LB↔nginx config (PROXY protocol, `externalTrafficPolicy: Local`, `use-forwarded-headers`) — **none of which is in this repo** (the Ingress has only `ssl-redirect`/`proxy-body-size`; no controller ConfigMap, no LB annotations). Two failure modes:

- **Collapse:** if the chain SNATs and nginx sets XFF to its TCP peer (LB/node IP), leftmost-XFF becomes a few LB/node IPs → the "per-IP" gate is effectively per-project. Weak, not a hole.
- **Forgery (security-critical):** if `use-forwarded-headers: true` is set anywhere (commonly needed to recover the client IP *through* a LB), nginx **appends** the client's XFF and the leftmost entry is client-controlled. Then `trust_proxy=true` makes **every** per-IP gate bypassable by header rotation — signup/email-spam, the OTP-verify brute-force from #233, token refresh. That's the exact bypass these gates exist to prevent, now on by default.

**Recommend confirming empirically** what the gateway actually receives before trusting the default: are `data_access_log` / audit client IPs distinct real client IPs, or all the same LB/pod IP? If real and unforgeable → the default is right; if forgeable → this is a one-flip regression across all gates.

**Hardening:** "leftmost XFF + trust the infra" is fragile (correct for exactly one overwriting hop). A trusted-hop-count / known-proxy-CIDR approach (rightmost-untrusted entry, or nginx's real-ip module setting one authoritative client IP the gateway reads) is robust to both extra hops and forgery. At minimum, document the required config (`use-forwarded-headers: false` **and** PROXY protocol or `externalTrafficPolicy: Local` on the Scaleway LB) as a hard precondition for this feature — its correctness is entirely that config.

### Minor
- Leftmost-XFF isn't `TrimSpace`d — fine today (leftmost has no leading space), but a future prepended hop with `", "` would yield a space-prefixed key; trivial to harden in the shared extractor.
- `SignupSigninPer5MinPerIP` correctly stays 8 (the bump to 30 remains gated on the #234/#235 email work).

Net: merge-quality code; I wouldn't rely on `trust_proxy=true` as the default until the ingress/LB XFF behavior is confirmed and ideally documented as a precondition.

---

Issue #238, opened 2026-06-27
[security][infra] Verify XFF behavior in Scaleway LB → nginx-ingress chain; harden ClientIPForProject to trusted-hop-count


STGime:
**Surfaced during #237 review.** The per-project `trust_proxy` knob ships with `false` as the default until two things happen: (1) we confirm what XFF actually looks like at the gateway in prod, and (2) we harden `ratelimit.ClientIPForProject` so its correctness doesn't rest on an infra assumption that isn't in this repo.

## Why this is open

`#237` wires the knob. The Eurobase nginx-ingress manifest (`deploy/k8s/ingress.yaml`) only sets `ssl-redirect` and `proxy-body-size` — no controller ConfigMap, no Scaleway LB annotations declaring PROXY-protocol or `externalTrafficPolicy: Local`. So what the gateway sees in `X-Forwarded-For` is determined by infra config that's invisible to this repo. Two failure modes for a default-true behaviour:

* **Collapse** — if SNAT happens somewhere and nginx writes XFF to its TCP peer (LB/node IP), leftmost-XFF lands on a few LB/node IPs across all real users. Per-IP gates degrade to weak per-cluster pseudo-keys. Not a hole; just wrong.
* **Forgery (security-critical)** — if `use-forwarded-headers: true` is set anywhere (commonly needed to recover the client IP through a load balancer), nginx **appends** the client-sent XFF and leftmost is **caller-controlled**. With `trust_proxy=true` the gateway then keys per-IP gates on attacker-rotated values → every per-IP rate limit becomes bypassable. Signup spam, OTP brute force (#233), token refresh — all the gates that exist to prevent abuse.

### 1. Empirical verification of the XFF chain
A small one-shot check from a pod that can reach the gateway:

```sh
# from inside the cluster, with curl bound to the gateway service:
curl -fsS https://api.eurobase.app/health \
  -H 'X-Forwarded-For: 1.2.3.4, 5.6.7.8'
# then inspect the gateway access log / data_access_log for that
# request's recorded client IP. Repeat from outside the cluster too.
```

Compare against `data_access_log` rows from the last hour (a real user-facing endpoint). The answer to **"what does `ratelimit.ClientIPForProject(r, true)` return in prod today?"** decides whether default-true is safe to flip.

Capture the answer in this issue's thread, then either land the default flip in a follow-up PR or move on to step 2.

### 2. Hardening: trusted-hop-count / known-proxy-CIDR
[...]

Issue #235, opened 2026-06-26
Per-project email sending: custom SMTP (BYO) + paid Scaleway-hosted email add-on


STGime:
## Why

Surfaced during #234 review. The per-project email quota (#227) enforces `EmailsPerHour` against the **single shared platform `EmailService`** — there is currently **no way for a project to send through its own sender**. That makes the platform email budget a hard ceiling: a project that outgrows it can only ask us to raise a number, and every project shares the platform sender's reputation/deliverability.

Supabase makes its low built-in-SMTP limit acceptable by pairing it with **"configure your own SMTP to lift it."** We need the same lever — and we can also turn it into revenue.

This issue covers two related deliverables: (1) BYO custom SMTP, free; (2) a paid managed email add-on we run on Scaleway.

## Part 1 — BYO custom SMTP (free tier lever)

Let a project configure its own SMTP / email provider; when set, auth + transactional emails for that project send through it and are **exempt from (or get a much higher) the platform `EmailsPerHour` quota**.

- [ ] Per-project email-sender config in `auth_config` (or a dedicated table): host, port, username, from-address, encryption. **Password/credential sealed at rest via the per-tenant vault key** (same pattern as #206 `env_vars` — `vault.SealForTenant`/`OpenForTenant`), never stored plaintext.
- [ ] `internal/email/EmailService` selects the project sender when configured, else the platform sender.
- [ ] Quota interaction: project-sender sends bypass the platform `quota:email:project:{id}` cap (the project owns its provider's limits) — or apply a generous separate ceiling. Keep the platform-sender path on the (raised, per #234) default.
- [ ] Console: "Email / SMTP" settings page — configure, send test email, show verified/last-error state.
- [ ] Verify-sender flow (send a test, surface SMTP errors) so a misconfigured SMTP fails loudly at setup, not silently at first signup.
- [ ] Sovereignty: warn if the configured SMTP host is a non-EU provider (consistent with `CLAUDE.md`).

## Part 2 — Paid Eurobase managed email add-on (Scaleway-hosted)

A first-party, EU-sovereign transactional-email add-on we sell, so projects get high deliverability without standing up their own SMTP — and it stays in-jurisdiction (Scaleway TEM, fr-par), unlike pointing BYO-SMTP at a US provider.

- [ ] Built on **Scaleway Transactional Email (TEM)** in `fr-par` (EU-sovereign; no US email vendor — consistent with the no-US-services rule).
[...]

PR #239, opened 2026-06-27
fix #230: rate-limits docs + at-the-limit integration tests

### LINKED ISSUE #230: [v1.3][rate-limits #F] Docs (rate-limits.md) + at-the-limit integration tests
Tracked under #224. Lands after #225–#229 so the docs reflect what shipped.

### Docs: `docs/compliance/rate-limits.md`
- One section per knob: name, default, what counts as "one request", where the counter resets, what the 429 looks like, how to read the corresponding Redis key from a pod.
- Table of defaults (mirrors the console UI; single source of truth).
- A worked example for IP forwarding: with `trust_proxy=false` what an attacker can/can't do; with `trust_proxy=true` what a legitimate edge can/can't do.
- "Debugging" section — `redis-cli` snippets for the auth limiter keys and the email/SMS hourly counter keys, all read-only.

### Integration tests: `internal/ratelimit/project_test.go`
At-the-limit + below-the-limit assertions for each of the five auth knobs:
- N requests pass; N+1 returns 429.
- After window expiry, next request passes.
- Per-project override takes effect on the next call (no need to wait for window expiry).
- Trust-proxy switch correctly changes the counter identifier.

### Email/SMS quota tests
- 2 emails go through; 3rd returns 429 with `Retry-After`.
- TEM client mock asserts `Send` was called twice, not three times.
- Same for SMS.

## Acceptance
- `go test ./internal/ratelimit/... -run Project -count=1` green.
- `docs/compliance/rate-limits.md` rendered and linked from the docs index.
- A reader of the doc can debug a 429 in production using only the snippets in the "Debugging" section.

STGime:
Closes #230 (umbrella #224).

## What lands

- **`docs/compliance/rate-limits.md`** — operator-facing reference. Defaults table (single source of truth, calls out the EmailsPerHour parked-pending-#235 reality), key-shape doc for both the auth keyspace and the quota keyspace, fail-open contract, IP-source / `trust_proxy` walkthrough with both failure modes spelled out (forgery vs collapse), Eurobase default rationale, precondition for the future flip (#238). `redis-cli` snippets for the three common questions: "is this IP currently capped?", "where did this project's hourly SMS budget go?", "force a counter reset". Override precedence at the bottom; related-issues at the end so the doc points outwards.
- **`internal/ratelimit/project_test.go`** — integration tests against a real Redis (skipped via setupAuthTestLimiter when one isn't available). Four cases:
  - `TestKnobs_AtTheLimit` (table-driven): for each per-IP knob (signup_signin, token_refresh, token_verify), N below-cap requests pass, (N+1)th returns 429 with Retry-After, and a second project on the same Redis + same identifier still gets through.
  - `TestKnob_WindowExpiry`: an at-cap counter unblocks once the limiter's window elapses.
  - `TestKnob_PerProjectOverrideTakesEffect`: tightening the limit mid-flight is reflected on the very next call without waiting for the window. Models the console-save UX.
  - `TestTrustProxy_ChangesCounterIdentifier`: at-cap-under-false + flip-to-true → next request passes because the identifier switched from TCP peer to leftmost XFF (different counter). The visible behaviour for the #228 knob.

The hourly-quota path coverage (CheckProjectHourlyQuota + ErrQuotaExceeded) ships with PR #234's own test file since those symbols don't exist on main until #234 merges.

`go test ./internal/ratelimit/... -count=1` green; all four new tests pass.

### DIFF
--- a/docs/compliance/rate-limits.md
+++ b/docs/compliance/rate-limits.md
@@ -1,0 +1,224 @@
+# Per-Project Rate Limits
+
+> Umbrella issue [#224](https://github.com/STGime/euroback/issues/224). Mirrors
+> Supabase's published Rate Limits page so the UX is familiar; the numbers
+> and the enforcement architecture are described below. Console UI:
+> **Auth → Rate Limits** tab on `/p/{id}/auth`.
+
+## What it is
+
+A small set of per-project knobs that decide how much load any one IP (or
+the whole project) can put on the platform's auth and notification
+pipelines. They exist to safeguard:
+
+1. **The platform**, against burst abuse from one project tanking
+   shared infrastructure (Scaleway TEM sender reputation, GatewayAPI
+   spend, the auth pool's RPS budget).
+2. **The project owner**, against runaway costs from a
+   misconfigured client or a leaked anon key.
+3. **End users**, against brute-force on signup, sign-in, token
+   refresh, and OTP verify.
+
+The numbers are tuned to give legitimate apps room to breathe and
+hostile traffic a wall to hit. Project owners can tighten or loosen
+each knob from the console; the underlying limiter uses the per-tenant
+shape so two projects on the same Redis can never collide.
+
+## Defaults
+
+These are the values returned by `DefaultRateLimits()` in
+`internal/tenant/auth_config.go`. Empty fields in the console form
+fall back to these. If you change one in code, update this table.
+
+| Knob | Default | Unit | Notes |
+|---|---|---|---|
+| `signup_signin_per_5min_per_ip` | **8** | requests / 5 min / IP | Held below Supabase's 30 until `emails_per_hour` enforcement lands (#235); the lower interim caps the verification-email amplification path. Bumps to 30 then. |
+| `token_refresh_per_5min_per_ip` | **150** | requests / 5 min / IP | High because legitimate SDK clients refresh proactively. |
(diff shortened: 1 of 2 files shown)


STGime (commented):
## Review — rate-limits docs + integration tests

**LGTM.** Strong closing piece for the #224 series.

- **The doc is honest and accurate.** Verified the most consequential claim against the code: `emails_per_hour` really is *defined-but-not-enforced* (`checkEmailQuota` has no call site on this base — parked behind #235), and the defaults table says exactly that, so it doesn't oversell. It also correctly covers SMS enforced-with-silent-skip, both key shapes (`auth:…` vs `quota:…`), fail-open, the `trust_proxy` forgery-vs-collapse trade-off, the #238 precondition for the default flip, override precedence, and read-only `redis-cli` debugging snippets. Genuinely useful operator reference.
- **Tests are solid integration coverage** — at-the-limit (table-driven per knob + per-project isolation), window expiry, override-takes-effect-next-call, and the `trust_proxy` identifier switch. Run against real Redis and pass; skip cleanly without it.

Nit (non-blocking): defaults now live in three places — backend `DefaultRateLimits()`, this doc table, and #240's `DEFAULT_RATE_LIMITS` TS const. Both carry "keep in sync" warnings, but given the signup default already flip-flopped (8→30→8), a small `/rate-limit-defaults` endpoint (or codegen) would eventually remove the drift risk.

---

PR #240, opened 2026-06-27
fix #229: console — Auth → Rate Limits tab

### LINKED ISSUE #229: [v1.3][rate-limits #E] Console: Auth → Rate Limits tab
Tracked under #224. Depends on #225, #226, #227, #228.

## What lands
A new "Rate Limits" tab on the existing Auth page (`console/src/routes/(app)/p/[id]/auth/+page.svelte`) — same tab pattern as today's "Settings" and "Email Templates" tabs.

### Fields
- **Rate limit for sending emails** — `emails_per_hour` — number, placeholder `2`
- **Rate limit for sending SMS messages** — `sms_per_hour` — number, placeholder `30`
- **Rate limit for token refreshes** — `token_refresh_per_5min_per_ip` — number, placeholder `150`, helper text "(N × 12 = N requests/hour)"
- **Rate limit for token verifications** — `token_verification_per_5min_per_ip` — number, placeholder `30`, helper text "(N × 12 = N requests/hour)"
- **Rate limit for sign-ups and sign-ins** — `signup_signin_per_5min_per_ip` — number, placeholder `30`, helper text "(N × 12 = N requests/hour)"
- **IP Address Forwarding** section heading + toggle for `trust_proxy` — copy: "Clients can forward end-user IP addresses to Auth for rate limiting when using server keys."

### Persistence
- `updateProject` in `console/src/lib/api.ts:401` already accepts an `auth_config` field. Extend the `AuthConfig` TS type with the new `rate_limits` sub-object; no method change.
- An empty number field saves `null` for that knob — the backend's `DefaultRateLimits` fallback (#A) handles the rest.
- Save button disabled until something changes (existing pattern on the Auth page).

## Acceptance
- Tab visible on `/p/{id}/auth`. Defaults shown as placeholders when the knob is unset.
- Edit two fields, save, reload → values persist.
- Clear a field, save → DB shows `null` for that knob; UI renders the placeholder default.
- Toggle the trust-proxy switch → `PATCH /v1/tenants/{id}` body includes `auth_config.rate_limits.trust_proxy: true|false`.

## Files
- `console/src/routes/(app)/p/[id]/auth/+page.svelte`
- `console/src/lib/api.ts` (type extension only)

STGime:
Closes #229 (umbrella #224). New tab on the existing Auth page mirroring the Supabase Rate Limits surface. 5 numeric inputs + 1 toggle, persists via existing updateProject. Honest per-knob copy (email parked pending #235, trust_proxy trade-off, silent-skip semantics for SMS). See commit body for design.

### DIFF
--- a/console/src/lib/api.ts
+++ b/console/src/lib/api.ts
@@ -37,7 +37,38 @@ export interface AuthConfig {
 	/** Browser origins (scheme://host[:port]) that may call this project's API.
 	 * Platform origins (eurobase.app, *.eurobase.app) are always allowed; this list is additive. */
 	cors_origins?: string[];
+	/** Per-project Rate Limits sub-object (#224 umbrella). Absent / null
+	 * means "use platform defaults". Each numeric field follows
+	 * zero-means-default on the backend (RateLimits.EffectiveRateLimits
+	 * merge), so the client can send 0 OR omit a field to reset that knob. */
+	rate_limits?: RateLimits;
 }
+
+/** Per-project rate-limit overrides. Numeric zeros / absent fields fall
+ * back to platform defaults on the backend. `trust_proxy` is *bool on
+ * the server so explicit `false` and "absent" are distinct (absent →
+ * use the platform default). */
+export interface RateLimits {
+	signup_signin_per_5min_per_ip?: number;
+	token_refresh_per_5min_per_ip?: number;
+	token_verification_per_5min_per_ip?: number;
+	emails_per_hour?: number;
+	sms_per_hour?: number;
+	trust_proxy?: boolean;
+}
+
+/** Platform-side defaults for the Rate Limits surface. Kept here so the
+ * console can show placeholder values that match the backend's
+ * `DefaultRateLimits()` without an extra API round-trip. If the backend
+ * defaults shift, update both sides in lock-step. */
+export const DEFAULT_RATE_LIMITS = {
+	signup_signin_per_5min_per_ip: 8,
+	token_refresh_per_5min_per_ip: 150,
+	token_verification_per_5min_per_ip: 30,
+	emails_per_hour: 2,
+	sms_per_hour: 30,
+	trust_proxy: false
+} as const;
 
 export interface PlatformProfile {
 	id: string;
--- a/console/src/routes/(app)/p/[id]/auth/+page.svelte
+++ b/console/src/routes/(app)/p/[id]/auth/+page.svelte
@@ -101,6 +101,16 @@
 	let rlTokenVerify = $state('');
 	let rlSignupSignin = $state('');
 	let rlTrustProxy = $state<boolean>(DEFAULT_RATE_LIMITS.trust_proxy);
+	// `rlTrustProxyTouched` distinguishes "the user clicked the toggle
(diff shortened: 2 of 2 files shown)


STGime (commented):
## Review — console Auth → Rate Limits tab

**Good, honest UI; one real finding that undermines the #237/#238 design.**

### Strengths
- Clean tab mirroring Supabase; numeric fields correctly omit-on-empty (empty → `undefined` → backend zero-means-default).
- Honest per-knob copy — email "saves but isn't applied yet (#235)", SMS silent-skip, `trust_proxy` off-by-default with the #238 note.
- Good a11y (`role="switch"` / `aria-checked`, labelled inputs); merges `rate_limits` into the existing `auth_config` rather than clobbering other fields. Hardcoded `DEFAULT_RATE_LIMITS` matches the backend (`signup_signin: 8`, `trust_proxy: false`).

### Finding — `trust_proxy` is always written as an explicit bool on save, defeating the `*bool` "absent = use default" design
`handleSaveRateLimits` always sends `trust_proxy: rlTrustProxy`, so the moment an owner saves this tab — even just to set an email number, even without touching the toggle — the stored config gets an **explicit `false`**. Per #237's `*bool` semantics, explicit-false means "stay false even if the platform default later flips to true" — which is the exact distinction #238 introduced so it *can* flip the default to `true` after verifying the XFF chain. So every project that opens and saves this page silently opts out of the future #238 flip. Harmless today (default is already false), but it strands early adopters on `false` the day #238 lands.

**Fix:** only include `trust_proxy` in the payload when the user actually changed it — track a "touched" flag (or make it tri-state default/on/off) and omit it otherwise, so it stays absent (nil → default) like the numeric fields already do.

### Minor
- Save button isn't disabled-until-dirty (the #229 acceptance asked for it, and it compounds the above — a no-op save still pins `trust_proxy: false`).
- `DEFAULT_RATE_LIMITS` is a hardcoded mirror of the backend defaults (drift risk — same note as on #239).

Net: solid UI, but I'd fix the `trust_proxy`-always-written behavior before #238 flips the default, otherwise it silently excludes any project that touched this page.

---

STGime (commented):
## Re-review (commit `d6d59f3`) — finding resolved ✅

The `trust_proxy`-always-written issue is fixed exactly as suggested:

- New `rlTrustProxyTouched` flag (default `false`), set `true` only in the toggle's `onclick`.
- Save now does `if (rlTrustProxyTouched) { rl.trust_proxy = rlTrustProxy }` — so `trust_proxy` is **only** persisted when the user explicitly flips it; otherwise it's omitted from the payload and stays absent (nil → platform default).
- `hydrateRateLimits` resets the flag to `false` after load/save, so a later unrelated save won't re-pin it.

Net effect: a project that opens and saves this tab without touching the toggle no longer gets an explicit `false`, so the `*bool` "absent = use default" contract holds and the future #238 default-flip to `true` will reach those projects. The two minors (Save not disabled-until-dirty; `DEFAULT_RATE_LIMITS` hardcoded mirror) remain but are now purely cosmetic — the no-op-save consequence that made the disabled-until-dirty point matter is gone.

LGTM. (Didn't run the console `npm` build here, but the change is a small, logically-sound Svelte edit.)

---

